PR #5, opened 2020-03-13
GdxRoot

lsizemore8:
Added applicable assets from the Gameplay Prototype, GDXRoot, and LoadingMode. 

Note: Game will still not run right now without GameCanvas. Also this will not work without the GameEngine PR request


BryanGraeser (approved):
Cool

PR #12, opened 2020-03-14
Level Controller & Level Model

BryanGraeser:
Mostly complete implementation of LevelController (waiting on AIController, PlayerModel, and EnemyModel to ensure correctly structured).

Framework implementation of LevelModel. All functions are present, but not all are implemented (waiting on AIController, PlayerModel, and EnemyModel to ensure correctly structured).

I also made the GameCanvas setCameraPosition function public as required (and added a javadoc comment for it), and cleaned up the comments crediting Walker White.


BryanGraeser:
Note I also still need to add player action functions to LevelController, just waiting on GameEngine to determine what exactly those need to be

BryanGraeser:
LevelModel implementation complete. Currently using tile size equivalent to the player's radius.

core/src/com/fallenflame/game/LevelController.java
```diff
@@ -1,5 +1,498 @@
 package com.fallenflame.game;
 
+import com.badlogic.gdx.math.*;
+import com.badlogic.gdx.physics.box2d.*;
+import com.badlogic.gdx.utils.*;
+import com.fallenflame.game.physics.obstacle.Obstacle;
+import com.fallenflame.game.physics.obstacle.ObstacleCanvas;
```
lsizemore8: This is unnecessary, we do not have an ObstacleCanvas

core/src/com/fallenflame/game/LevelController.java
```diff
@@ -1,5 +1,498 @@
...
+        flares = new LinkedList<>();
+        levelModel = new LevelModel();
+    }
+
+    /**
+     * Lays out the game geography and enemies from the given JSON file
+     *
+     * @param levelFormat	the JSON tree defining the level
+     */
+    public void populate(JsonValue levelFormat) {
```
lsizemore8: For consistency's sake (also super picky), do we want to call our level JSON levelJson or levelFormat? I made GameEngine call it levelJson cause the other JSON files had that type of naming

core/src/com/fallenflame/game/LevelController.java
```diff
@@ -1,5 +1,498 @@
...
+    protected float physicsTimeLeft;
+
+    /**
+     * Returns the bounding rectangle for the physics world
+     *
+     * The size of the rectangle is in physics, coordinates, not screen coordinates
+     *
+     * @return the bounding rectangle for the physics world
+     */
+    public Rectangle getBounds() { return bounds; }
```
lsizemore8, line 83: Not absolutely sure here, but I think we'll need a setBounds too if we want a resizeable canvas (or any sized canvas that isn't set in the level JSON

core/src/com/fallenflame/game/LevelController.java
```diff
@@ -1,5 +1,498 @@
...
+     * of how tightly coupled everything is.
+     *
+     * @param dt the time passed since the last frame
+     */
+    public void update(float dt) {
+        // Update player (and update levelModel) and exit
+        levelModel.removePlayer(player);
+        player.update(dt);
+        levelModel.placePlayer(player);
+        exit.update(dt); // TODO: Necessary?
```
lsizemore8: Probably not, can probably just check if we at the exit bounds or not.

core/src/com/fallenflame/game/LevelController.java
BryanGraeser: Nice catch, thanks!

lsizemore8 (changes requested):
There are a couple functions GameEngine needs that aren't here 
1. movePlayer-> we talked about doing all movement of characters in levelController, so gameEngine needs a way to send levelController an angle and a force and for you to basically call applyForce/setAngle/setMovement on the player
2. What is called to expand the light radius? 

Past that, it was a couple small fixes/naming consistencies.

core/src/com/fallenflame/game/LevelController.java
```diff
@@ -1,5 +1,498 @@
...
+        // Update lights
+        lightController.updateLights(player, flares, enemies);
+    }
+
+    /**
+     * Launch a flare from the player towards the mouse position based on preset flareJSON data
+     *
+     * @param mousePosition Position of mouse when flare launched
+     */
+    public void createFlare(float[] mousePosition){
```
lsizemore8: Why are you using a float array as opposed to a Vector2?

BryanGraeser: true, I can switch it

lsizemore8 (approved):
Looks good!

PR #13, opened 2020-03-14
Start player and enemy model

foxhatleo:
A few things:

1. I have included textures of player model and enemy model from gameplay prototype. But it needs to be linked in JsonAssetManager.
2. Right now enemy has practically the same behaviour as player besides max speed and texture so enemy is a subclass of player, although we can revisit that.


BryanGraeser:
Are we sure we want EnemyModel to extend PlayerModel? They do share a lot of functions and having EnemyModel extend PlayerModel will make it easier when we update/change any of those shared functions, but it also means that when we add new features to PlayerModel later (such as cooldown for abilities and such) that EnemyModel will inherit these features as well--I'm not sure that this makes sense. If anything, we can have a a model that Enemy and Player both extend.

BryanGraeser (changes requested):
[See my earlier comment for context] For example, I needed to add an incrementLightRadius function to playerModel (to allow level and input controller to not manage the light radius), but now enemymodel also inherits this functions which doesn't make sense. Can you change it so EnemyModel does not extend PlayerModel. You can make them both extend some other more general CharacterModel if you'd like.

BryanGraeser:
Okay, so I've refactored EnemyModel and PlayerModel to extend an abstract class CharacterModel. EnemyModel and PlayerModel initialize rn is hard-coded which may be better in the short-term for testing purposes, but will definitely need to be changed to dynamically initialize from the JSON. I have filed a github issue regarding this so we know to come back to it.

BryanGraeser (dismissed):

BryanGraeser (approved):
gtg

PR #16, opened 2020-03-21
Flare Model

BryanGraeser:
Flare Model implementation and corresponding changes to level controller. I wrote it with film strip, but we can change it to a static texture to make it simpler.


foxhatleo (approved):
Looks good.

PR #17, opened 2020-03-21
AI controller

ronnie-dumesh:
Added compatibility for AIController with APIs for LevelModel, FlareModel, PlayerModel, and EnemyModel. Enemies should now be able to chase enemies if they are spotted and to approach the last-known location of enemies and to approach flares if they are in the enemies' ranges. Fiddling with constants must be done.


core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+    /** The radius from which a monster can chase a player */
+    private static final int CHASE_DIST = 1000;
+    /** The radius from which a monster can attack a player */
+    private static final int ATTACK_DIST = 1000;
+    /** The radius from which an enemy could have considered to have finished its investigation
+     * of a flare or of a player's last-known location*/
+    private static final int REACHED_INVESTIGATE = 1000;
+
+    // Instance Attributes
+    /** The ship being controlled by this AIController */
```
BryanGraeser: A few of the comments still reference "ship". I'm assuming from lab2, can you do a quick find&replace for these?

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+    /** The player*/
+    private PlayerModel player;
+    /** The flares that the enemy will investigate */
+    private List<FlareModel> flares;
+    /** The flares that have been investigated by the enemy */
+    private HashSet<FlareModel> investigatedFlares;
+    /** The last-known position of the player*/
+    private Vector2 investigationPosition;
+    /** The ship's next action. */
+    private Action move; // A ControlCode
```
BryanGraeser, line 62: We may want to change this to "action" not move for when enemies have other kinds of actions later

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+    /** The enemy's current state*/
+    private FSMState state;
+    /** The player*/
+    private PlayerModel player;
+    /** The flares that the enemy will investigate */
+    private List<FlareModel> flares;
+    /** The flares that have been investigated by the enemy */
+    private HashSet<FlareModel> investigatedFlares;
+    /** The last-known position of the player*/
+    private Vector2 investigationPosition;
```
BryanGraeser: Isn't this also potentially the location of a flare? Not just the last-known player position?

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+
+                break;
+
+            case CHASE:
+                enemy.setActivated(true);
+
+                if(rand_int < 30 || !withinChase()){ //random chance of quitting chase
+                    state = FSMState.INVESTIAGE;
+                    investigationPosition = new Vector2(player.getX(), player.getY());
+                }
```
BryanGraeser: Do we want this? I thought the enemies chased the player until they were out of visibility range (and then they go to the player's last known location)

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+                break;
+
+            case ATTACK:
+                if(!withinAttack()){
+                    state = FSMState.CHASE;
+                }
+
+                break;
+
+            case INVESTIAGE: //investigation position must not be null
```
BryanGraeser: Spelling error

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+                break;
+        }
+    }
+
+    /**
+     * Mark all desirable tiles to move to.
+     *
+     * This method implements pathfinding through the use of goal tiles.
+     * It searches for all desirable tiles to move to (there may be more than
+     * one), and marks each one as a goal. Then, the pathfinding method
```
BryanGraeser: Can there be more than one goal tile?

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+     * getMoveAlongPathToGoalTile() moves the ship towards the closest one.
+     *
+     * POSTCONDITION: There is guaranteed to be at least one goal tile
+     * when completed.
```
BryanGraeser, line 203: This postcondition is no longer true. This is fine, but it shouldn't be commented here if it no longer applies

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+
+                break;
+
+            case CHASE:
+                sx = level.screenToTile(playerX);
+                sy = level.screenToTile(playerY);
+                if(level.getSafe(sx, sy)){
+                    enemy.setGoal(playerX, playerY);
+                    setGoal = true;
+                }
```
BryanGraeser: We don't need to check that this is safe. The enemy wants to go here regardless.

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+                break;
+
+            case ATTACK:
+                sx = level.screenToTile(playerX);
+                sy = level.screenToTile(playerY);
+                if(level.getSafe(sx, sy)){
+                    enemy.setGoal(playerX, playerY);
+                    setGoal = true;
+                }
```
BryanGraeser: Same as above comment

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+                break;
+
+            case INVESTIAGE: //investigationPosition must not be null
```
BryanGraeser: spelling error

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+        }
+    }
+
+    /**
+     * Returns a movement direction that moves towards a goal tile.
+     *
+     * The value returned should be a control code.  See PlayerController
+     * for more information on how to use control codes.
+     *
+     * @return a movement direction that moves towards a goal tile.
```
BryanGraeser: or returns NO_ACTION

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+        HashSet<Coordinate> visited = new HashSet<>();
+
+        while (!queue.isEmpty()) {
+
+            c = queue.poll();
+            if (enemy.isGoal(c.getX(), c.getY())) {
+                break;
+            }
+
+            if (visited.contains(c)) {
```
BryanGraeser: I think you mean `if (!visited.contains(c))`

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+    // Constants
+    /** The radius from which a monster will notice a flare and approach it*/
+    private static final int FLARE_DETECTION_RADIUS = 1000;
+    /** The radius from which a monster can chase a player */
+    private static final int CHASE_DIST = 1000;
+    /** The radius from which a monster can attack a player */
+    private static final int ATTACK_DIST = 1000;
+    /** The radius from which an enemy could have considered to have finished its investigation
+     * of a flare or of a player's last-known location*/
+    private static final int REACHED_INVESTIGATE = 1000;
```
BryanGraeser, line 43: Shouldn't this value be much smaller than the other constants? Otherwise the enemy will immediately think it has reached investigate position

core/src/com/fallenflame/game/AIController.java
```diff
@@ -1,5 +1,411 @@
...
+                break;
+
+            case INVESTIAGE: //investigationPosition must not be null
+                float investX = investigationPosition.x;
+                float investY = investigationPosition.y;
+                sx = level.screenToTile(playerX);
+                sy = level.screenToTile(playerY);
+                if(level.getSafe(sx, sy)){
+                    enemy.setGoal(investX, investY);
+                }
```
BryanGraeser: Probably won't be safe bc flare sticks to wall. Again I don't think we should check safety of goal locations, that should just be for movement

core/src/com/fallenflame/game/AIController.java
BryanGraeser: Also how is setting player's position as investigate position and then investigating different than chasing them?

BryanGraeser (changes requested):
Okay I left some specific comments in the code, so if you could review and address those, that'd be great. But I also have some more general questions/issues: 
1. **FSM Stuff.** What is the purpose of both a Chase and Attack state? From what I can see, they set the same goal tiles, so why do we need both? Further, there seems to be no transition out of the chase or attack state back to wander or spawn. This means once the AI starts chasing the player they will never stop, which is not what we want. I drew out the FSM as I interpreted from your code and I can attach that to this comment, but that is how I found these issues. Can you draw out the FSM as you intend it and send it to me? I think that would make this more clear so we can resolve these issues.
2. **Player Light Radius** Nowhere in here do you take into account the player's light radius. You have the enemy chase the player when they are within a certain, static range regardless of the player's light radius which is not what we want. You need to check the player's light radius and **if and only if** the enemy is within the light radius are they activated.

It seems like a lot of these issues are related to left-overs from lab2, so I think we just need to focus on adapting this more to our game instead of trying to use what they had.

BryanGraeser:
Here is the current FSM you have as I see it. If you could please draw out the version you want/are envisioning before you implement the changes, that'd be much appreciated

core/src/com/fallenflame/game/AIController.java
ronnie-dumesh: Yes, changed docstring to reflect this

core/src/com/fallenflame/game/AIController.java
ronnie-dumesh: The reason it's called move is exactly for that reason. It calculates a direction to move every time, before determining whether another action is appropriote. Here is the code that uses move, it's to essentially avoid calculating the move every tick. 

        if ((randomID + ticks) % 10 == 0) {
            changeStateIfApplicable();

            markGoalTiles();
            move = getMoveAlongPathToGoalTile();
        }

        Action action = move;
        return action;

core/src/com/fallenflame/game/AIController.java
ronnie-dumesh: Yes, I just have these as placeholders.

core/src/com/fallenflame/game/AIController.java
ronnie-dumesh: investigationPosition is for the INVESTIGATE state. Chasing updates every single time an enemy is chasing a player, but investigationPosition is set when the player transitions into INVESTIGATE. Once an enemy is in INVESTIGATE, it receives no further information about the player's position. This is how I implemented the enemy going to the player's last-known location

ronnie-dumesh:
FSM discussed over slack. Now FSM made up of IDLE (in place of WANDER), CHASE, ATTACK, INVESTIGATE states. SPAWN state was removed

I also accounted for the light radius of the player in the method withinChase(). I think withinChase() and withinAttack() will need to be fine-tuned as we experiment with gameplay, but that will require us to test the program and make design decisions.

core/src/com/fallenflame/game/AIController.java
ronnie-dumesh: As for the random quitting chasing, I think that's a discussion to be had but I've removed it from now. It's basically to prevent the enemies from being near-certain kill-bots and to give the player a chance at evading them in a chase, especially being that enemies are faster than players

core/src/com/fallenflame/game/AIController.java
ronnie-dumesh: currently, we assumed no, changed docstring to reflect

core/src/com/fallenflame/game/AIController.java
ronnie-dumesh: Why is this not true? If an enemy is staying in the same place it's goal tile is its current position

core/src/com/fallenflame/game/AIController.java
ronnie-dumesh: It applies here as I'm picking a completely random position on the map, it may not always be safe.

ronnie-dumesh:
I need to check how having potentially unsafe goals may affect BFS, I don't think it will, but still will be important to test.

core/src/com/fallenflame/game/AIController.java
BryanGraeser: Ah you're right. I thought clearGoal() cleared all goals for the enemy, didn't realize it set the goal as the enemy's position

core/src/com/fallenflame/game/AIController.java
BryanGraeser: But what if the movement is NO_ACTION? For example, before the enemy is activated

ronnie-dumesh: move will be NO_ACTION in that case.

ronnie-dumesh: move will be NO_ACTION in that case.

BryanGraeser (approved):
Looks gtg to me now so you can go ahead and merge. 
Heads up: LevelModel is encountering some out-of-bounds issues currently because our coordinate system is a little funky and different than b2lights rn. I am working on fixing this issue, but temporarily levelmodel is not initializing any world objects so it will basically see the whole world as safe.

PR #19, opened 2020-03-22
Running version

BryanGraeser:
Myriad of bug fixes and filling in JSON. Highlights:
- change Java version to 1.8
- flare, player, exit, and general lighting json data added to level.json
- add light radius to flaremodel
- update function calls to correct name in GameEngine
- create CanvasBound rectangle in GameEngine which was previously not created leading to nullpointer
- update level controller to only dispose if populated
- made level controller a contact listener for the world
- comment out AIController dependent code
- light controller changed initialization ordering so that pointsource was not trying to use a rayhandler that did not yet exist

This is still not yet a runnable version, but it launches a level without crashing thanks to the above so that's progress. Currently is just a black screen when the level is launched.


core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -210,15 +212,15 @@ public void dispose() {
...
+        setIsSuccess(false);
+        setIsFailed(false);
+        // countdown = -1; TODO Laura: what is this for?
 
         // Reload the json each time
         String currentLevelPath = "jsons/" + saveJson.getString("current");
         levelJson = jsonReader.parse(Gdx.files.internal("jsons/level.json"));
         level.populate(levelJson);
-        level.getWorld().setContactListener(this);
+        level.getWorld().setContactListener(level); //TODO Laura: I switched this to level
```
lsizemore8, line 223: Yeah good move

core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -210,15 +212,15 @@ public void dispose() {
     public void reset() {
         level.dispose();
 
-        setComplete(false);
-        setFailure(false);
-        countdown = -1;
+        setIsSuccess(false);
+        setIsFailed(false);
+        // countdown = -1; TODO Laura: what is this for?
```
lsizemore8, line 217: This one added time to put a "you died" or a "you won" message before it reset. I wasn't sure if I wanted to keep it in or not, but looking at this more this seems necessary

lsizemore8: I'll add support in a separate branch/pr

lsizemore8 (changes requested):
Small changes: 
1. The player-walking.png and enemy-walking.png should be in the textures folder. 
2. The assets.json needs to be updated with the correct file name.
3. walking.png should be removed.  
4. We need to add a level background to the game once light is fixed
5. The screen size is way too small, it should be the same as the prototype for now until we can figure out how to do resizing. 
5. From what I can tell, the only thing being drawn on the black screen is the player, as only one thing popped up when I did the debug mode. This may have to wait to be fixed until light is working, but that is a problem rn. 

Also. Do we want the countdown feature? All it theoretically does is make the game wait a tad before restarting. When I played the gameplay prototype that countdown seemed way too long, but idk without testing. I'll try to play around with the gameplay prototype today to see and can update this pull request if it's needed. 

I'll do a second read over/testing after the initial fixes are made to double check.

BryanGraeser:
PR comments addressed except for:
4. We need to add a level background to the game once light is fixed
and 6. From what I can tell, the only thing being drawn on the black screen is the player, as only one thing popped up when I did the debug mode. This may have to wait to be fixed until light is working, but that is a problem rn.

We probably want some sort of cooldown so the player has time to read the message and realize they died, but it's not a priority.

core/assets/jsons/level.json
```diff
@@ -2,10 +2,49 @@
...
+    "bodytype":    	"dynamic",
+    "density":        1.0,
+    "friction":       0.0,
+    "restitution":    0.0,
+    "force": 		 80.0,
+    "damping": 		 10.0,
+    "maxspeed": 	  5.0,
+    "startframe":       0,
+    "walklimit":		4,
+    "texture":     	"dude",
```
lsizemore8: This needs to be "player" now, not "dude"

lsizemore8 (approved):
Yay!

PR #21, opened 2020-03-25
Fix for Physics/Player Movement

BryanGraeser:
Fixed physics issue by incorporating world.step.
Also added walls so that once moving camera is working, we can still tell the player is moving (note: moving camera is not working rn).

Finally LevelModel is having lots of indexoutofbounds issues that I will look into so I've commented out problematic levelmodel calls for now.


lsizemore8 (approved):
Looks good

PR #25, opened 2020-03-25
Wall model bug fix: render error and static body

### LINKED ISSUE #23: Walls issues
Issues observed so far:
- placement of walls seems off; is different from b2lights which is not necessarily bad, but @foxhatleo is looking into correct placement
- walls bounce when pushed by player

foxhatleo:
Resolves #23 

This PR make two changes:
- Make wall static so it does not move when something runs into it
- Fix render error so placement of wall is correct


BryanGraeser (approved):
Looks gtg. But the wall placement still does not line up with b2lights demo

PR #26, opened 2020-03-25
Player cam

stallop:
Made camera correctly follow player, issue was just not correctly using our scale for x and y.


BryanGraeser (approved):
gtg

PR #27, opened 2020-03-26
AI integration

BryanGraeser:
This integrates AI Controller into level controller. Currently LevelModel always returns true for isSafe (to prevent bugs that crash the application). This will be fixed later, but I want to merge now so @ronnie-dumesh can test AIController FSM and movement in a safe environment.


ronnie-dumesh (approved):
Looks good to me!

Issue #14, opened 2020-03-20
Testing Scroll Wheel


lsizemore8:
Checking the scroll wheel amount was done in a slightly strange way, so this'll have to be tested when everything is set up

foxhatleo:
I have added a temp alternative in light-controller branch that uses comma and period as an alternative to scroll wheel because a) this needs to work to test player light, and b) I don't have a mouse at the moment, just my touchpad. Maybe we can talk about preserving this alternative or some alternative way for people who don't have a mouse wheel?

lsizemore8:
I'm okay with keeping it as an alternative, but that is a discussion for everyone. I think I just figured out how to get the scrolling working without making a real weird architecture choice, so that should be fixed in a tad, so we should be able to have both.

PR #31, opened 2020-03-26
Light controller fix

foxhatleo:
This PR fixes player light. It should work as expected now. It does/fixes a few things.

- Player light now works, and light camera (which is independent from canvas camera) follows player too.
- I have not made things outside the light pitch dark just yet for debug purposes. However, that is just one single line of code that can be done at any time.
- Added a mouse wheel alternative of comma and period to serve as a temp fix to #14 
- **NOT TESTED** on enemies yet.


BryanGraeser (approved):
gtg

PR #32, opened 2020-03-26
Scroll wheel changes light radius

lsizemore8:
Added support to cause the scroll wheel to change the light radius.

Other note, this does change the architecture spec some, as this was the literal only way to do it without LoadingMode needing GameEngine. 

Third note: This does not have XBox Support, and the comma/period fix is still in the code in case we want it, although it will not work right now.

### DIFF
--- a/core/assets/jsons/level.json
+++ b/core/assets/jsons/level.json
@@ -19,7 +19,7 @@
     "debugopacity":	192
   },
   "exit":             {
-    "pos":            [ 1.5, 10.35],
+    "pos":            [ 1.5, 5.35],
     "size":           [ 1.25, 1.25],
     "bodytype":    	  "static",
     "density":  	  0.0,
--- a/core/assets/jsons/level.json
+++ b/core/assets/jsons/level.json
@@ -19,7 +19,7 @@
     "debugopacity":	192
   },
   "exit":             {
-    "pos":            [ 1.5, 5.35],
+    "pos":            [ 1.5, 10.35],
     "size":           [ 1.25, 1.25],
     "bodytype":    	  "static",
     "density":  	  0.0,
--- a/core/src/com/fallenflame/game/GameEngine.java
+++ b/core/src/com/fallenflame/game/GameEngine.java
@@ -268,9 +268,6 @@ public class GameEngine implements Screen {
         if (input.didFlare()) {
             level.createFlare(input.getMousePosition());
         }
-        if(input.didLight()){
-            level.lightFromPlayer(input.getLightRadius());
-        }
         // Rotate the avatar to face the direction of movement
         tempAngle.set(input.getHorizontal(),input.getVertical());
         float angle = 0;
@@ -283,6 +280,14 @@ public class GameEngine implements Screen {
         level.update(delta);
         isSuccess = level.getLevelState() == LevelController.LevelState.WIN;
         isFailed = level.getLevelState() == LevelController.LevelState.LOSS;
+    }
+
+    /**Increments the player's light radius
+     * @param amount, which represents the amount to increment the light radius*/
+
+    public void lightFromPlayer(float amount){
+        System.out.println("amount is: " + amount);
+        level.lightFromPlayer(amount);
     }
 
     /**
--- a/core/src/com/fallenflame/game/GameEngine.java
+++ b/core/src/com/fallenflame/game/GameEngine.java
@@ -220,7 +220,7 @@ public class GameEngine implements Screen {
         String currentLevelPath = "jsons/" + saveJson.getString("current");
         levelJson = jsonReader.parse(Gdx.files.internal("jsons/level.json"));
         level.populate(levelJson);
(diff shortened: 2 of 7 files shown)


core/assets/jsons/level.json
```diff
@@ -19,7 +19,7 @@
     "debugopacity":	192
   },
   "exit":             {
-    "pos":            [ 1.5, 10.35],
+    "pos":            [ 1.5, 5.35],
```
BryanGraeser: Why?

core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -285,6 +282,14 @@ public void update(float delta) {
         isFailed = level.getLevelState() == LevelController.LevelState.LOSS;
     }
 
+    /**Increments the player's light radius
+     * @param amount, which represents the amount to increment the light radius*/
+
+    public void lightFromPlayer(float amount){
+        System.out.println("amount is: " + amount);
```
BryanGraeser: Remove this println

core/src/com/fallenflame/game/LightInputProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+        return false;
+    }
+
+    public boolean touchDragged (int x, int y, int pointer) {
+        return false;
+    }
+
+    public boolean mouseMoved (int x, int y) {
+        return false;
+    }
```
BryanGraeser: Why are these all always false?

core/src/com/fallenflame/game/LightInputProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+       if(amount == 1){
+           gameEngine.lightFromPlayer(-1.0f);
+       }
+       if(amount == -1){
+           gameEngine.lightFromPlayer(1.0f);
+       }
+
+       return true;
+    }
+}
```
BryanGraeser: Can you add comments to this file?

BryanGraeser (changes requested):

BryanGraeser:
Additionally, can you explain why loadingMode is involved and why a separate LightInputProcessor is required? I am sure you have your reasons, but there are no comments explaining why.

core/assets/jsons/level.json
lsizemore8: That was so I could test where the exit was when it was originally broken. Forgot I had changed it, can change back

core/src/com/fallenflame/game/LightInputProcessor.java
lsizemore8: They need to be false so they don't do anything. Basically for these functions, the code defaults to what the LoadingMode InputProcessor does, unless it is the scrolled function. If you look in LoadingMode, the scrolled function returns false. It basically says "we don't use another listener" when those events happen for it to be false.

core/src/com/fallenflame/game/LightInputProcessor.java
```diff
@@ -0,0 +1,90 @@
...
+ * that LoadingMode needs versus event-driven events that the main game needs*/
+
+public class LightInputProcessor implements InputProcessor {
+   private GameEngine gameEngine;
+
+   public LightInputProcessor(GameEngine g){
+       gameEngine = g;
+   }
+   /**Below are functions that are required to be implemented by InputProcessor. All return false if unused to indicate
+    * that the event was not handled by this inputProcessor. All return that return false here
```
BryanGraeser: "All that return false here" very minor typo

BryanGraeser (commented):
Ok, these comments are very helpful, thanks! Couple questions and then I think it's gtg:

1. Why does scrolling need to be handled as event-driven as opposed to polling? Is it because detection of the 'scroll' is so instantaneous polling would not pick it up?
2. Will adding event-driven input have any impact on performance or the game loop?

core/src/com/fallenflame/game/LightInputProcessor.java
lsizemore8: lmao oops

lsizemore8:
> Ok, these comments are very helpful, thanks! Couple questions and then I think it's gtg:
> 
> 1. Why does scrolling need to be handled as event-driven as opposed to polling? Is it because detection of the 'scroll' is so instantaneous polling would not pick it up?
> 2. Will adding event-driven input have any impact on performance or the game loop?

The reason this is event-driven is because there is no support for polling under Gdx.Input like the other events. InputController gets away with polling by using "isKeyJustPressed" or "isButtonPressed" from the Input class, but the Input class has no such support for scrolling. When I was searching it up, this seemed to be the only way to have any mouse scrolling detection. 

This shouldn't have much performance issues, if any, given that most events won't do anything, this has an O(1) time complexity, and there's already event driven code occurring whenever a key is moved due to LoadingMode. 

I'll add this in comments as well.

BryanGraeser (approved):
GTG!

PR #33, opened 2020-03-26
Character JSON Initialization

BryanGraeser:
Changed enemy and player models to json initialization as opposed to hardcoding. This initialization occurs through CharacterModel. I removed the hardcoded functions that were originally pushed as well.

To understand the necessary JSON values required for character models look a the player json in fallen-flame or the json in b2lights demo.


ronnie-dumesh (approved):
Everything looks good, much cleaner too

PR #34, opened 2020-03-26
Concurrent access lightController fix

BryanGraeser:
Issue is for this code:
```
enemyLights.keySet().stream().filter(i -> !enemies.contains(i) || !i.getActivated()).forEach(i -> {
            PointSource f = enemyLights.get(i);
            f.setActive(false);
            f.dispose();
            enemyLights.remove(i);
        });
```
You can't stream/filter over enemyLights.keySet() while simultaneously changing the size of enemyLights.keySet() (via `enemyLights.remove(i);`).

I can rewrote this to use an iterator with appropriate removal to not cause a concurrent access issue

### DIFF
--- a/core/src/com/fallenflame/game/CharacterModel.java
+++ b/core/src/com/fallenflame/game/CharacterModel.java
@@ -168,11 +168,6 @@ public abstract class CharacterModel extends WheelObstacle {
         walkLimit = value;
     }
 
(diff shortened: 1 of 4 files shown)


ronnie-dumesh (approved):
Good solution! I tested and the program does not crash because of this anymore. Approved.

PR #37, opened 2020-03-27
Reset Handling

BryanGraeser:
I think this is how we should handle reset. Dispose the level and then create a new one. This will handle the lightController issue too because a new lightController will be created so I removed the fix Leo added before.


lsizemore8 (approved):
If you think this won't cause performance issues then I'm okay with it.

PR #38, opened 2020-03-27
Win Condition

lsizemore8:
Win Condition correctly displays the word "victory" above the goal and resets after a brief period of time. 

Note: because of the different coordinate systems for gameCanvas, victory has to be drawn in what the game sees as the upper left corner, despite the camera centering on that position.

### DIFF
--- a/core/src/com/fallenflame/game/GameEngine.java
+++ b/core/src/com/fallenflame/game/GameEngine.java
@@ -220,7 +220,7 @@ public class GameEngine implements Screen {
         String currentLevelPath = "jsons/" + saveJson.getString("current");
         levelJson = jsonReader.parse(Gdx.files.internal("jsons/level.json"));
         level.populate(levelJson);
-        level.getWorld().setContactListener(level); //TODO Laura: I switched this to level
+        level.getWorld().setContactListener(level);
     }
 
     /**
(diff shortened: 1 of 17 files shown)


core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -253,6 +264,11 @@ public boolean preUpdate(float dt) {
             listener.exitScreen(this, EXIT_QUIT);
             return false;
         }
+        else if (countdown > 0) {
+            countdown--;
+        } else if (countdown == 0) {
+            reset();
+        }
```
BryanGraeser, line 273: Why are we resetting if countdown is reached?

BryanGraeser (commented):
Why are you changing InputController in this PR?

I shouldn't be, that's strange. Likely a merge conflict gone wrong

core/src/com/fallenflame/game/GameEngine.java
lsizemore8: Countdown is -1 until we win or lose. It's used purely to show "Victory" or "You Died" for a certain amount of time (ie until countdown reaches 0). Once it reaches 0, reset

BryanGraeser:
gtg

BryanGraeser (approved):
gtg

PR #39, opened 2020-03-27
Exterior walls

foxhatleo:
This PR updates level.json by adding four extra walls forming a group of "exterior walls". On the programming side, the exterior wall is implemented differently than the gameplay prototype: There is no ExteriorWall model here, just one single box-shaped wall model so need four of them to make exterior.


BryanGraeser (approved):
gtg

PR #40, opened 2020-03-27
New updateLights() for light controller 

foxhatleo:
A few changes to `updateLights()` that solves a few issues:

- Prevents concurrent access, I think, so no more error on that end hopefully
- Adds an important step to keep asking enemies and flares for light radius every frame and render it accordingly
- Because the light update process for enemies and flares are similar, they have been abstracted out
- To allow this abstraction an interface called "`ILightRadius`" is created that requires implementation of "`getLightRadius():float`". This is necessary because two things need to implement it: "CharacterModel (EnemyModel and PlayerModel)" as well as "FlareModel".


BryanGraeser (approved):
gtg

PR #41, opened 2020-03-27
Fix diagonal speed problem

### LINKED ISSUE #36: Player moves faster diagonally
This is something easy to fix, just recording it so I don't forget.

foxhatleo:
Resolves #36


BryanGraeser (approved):
gtg

PR #46, opened 2020-03-27
Working Enemy AI

### LINKED ISSUE #30: Search-algorithm memory error
Possible to see in add-enemies branch 

In my BFS algorithm, I have Hashset<Coordinate> visited. When I go through the BFS, I add the position of the visited tiles to visited. When I run the game, it instantly freezes and crashes after some time. (You have to let the game crash on its own to reproduce the error, not close it). Make sure line 182 of LightController is commented out as well.

I think this is simply an unavoidable consequence of using a HashSet here. How Lab2 implemented visited tiles was through "marking". See getMoveAlongPathToGoalTile() description. http://www.cs.cornell.edu/courses/cs3152/2020sp/labs/program2/

We can implement this "marking" functionality in LevelModel and hopefully avoid this issue. Perhaps I can also try your BFS algorithm here as it might be with my implementation. Being that we're probably going to switch to an A*-type search algorithm anyway, we might want to have an option that allows easy transition to that.

Exception in thread "LWJGL Application" com.badlogic.gdx.utils.GdxRuntimeException: java.lang.OutOfMemoryError: Java heap space
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:135)
Caused by: java.lang.OutOfMemoryError: Java heap space
	at java.util.HashMap.resize(HashMap.java:704)
	at java.util.HashMap.putVal(HashMap.java:663)
	at java.util.HashMap.put(HashMap.java:612)
	at java.util.HashSet.add(HashSet.java:220)
	at com.fallenflame.game.AIController.getMoveAlongPathToGoalTile(AIController.java:272)
	at com.fallenflame.game.AIController.getAction(AIController.java:106)
	at com.fallenflame.game.LevelController.update(LevelController.java:382)
	at com.fallenflame.game.GameEngine.update(GameEngine.java:283)
	at com.fallenflame.game.GameEngine.render(GameEngine.java:337)
	at com.badlogic.gdx.Game.render(Game.java:46)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:233)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:128)

### LINKED ISSUE #48: Enemies glowing before activated
This is currently an issue on branch `fix-enemy-ai` will revisit if still an issue once merged into master, but wanted to catalogue it.

BryanGraeser:
This is a semi-working AI with pathfinding that generally works, but has some edge cases that need to be ironed out. I wrote my own AIController for this branch based off parts of Ronnie's AIController. I have simplified the FSM so we can get something working at a more basic level before adding complicated functionality and I am also using my BFS from lab2.

This PR is not complete, but I am creating a draft PR to increase visibility.


foxhatleo:
Resolves #48

BryanGraeser:
One known bug I am currently looking into is that the enemy will sometimes get stuck in a Left-Right infinite movement pattern.

BryanGraeser:
Increasing tile size seems to solve Left-Right issue, but introduces a new issue in that if the player is centered on an "unsafe" tile (even though the tile is only partially unsafe), then the enemy will not target them.
image
**[enemy does not move in above frame bc player is in "unsafe tile"**

BryanGraeser:
If you would like to mess around with `TILE_SIZE`, it is a constant in LevelModel

BryanGraeser:
Created fix where enemy can set an unsafe tile as their goal (may result in inaccessible goals if enemy must travel through more than one unsafe tile to reach player, but this should be impossible in **most** scenarios -- need to be careful about small pathways.

BryanGraeser:
**Note: increasing the enemies' speed later which we have discussed may re-introduce the left-right issue.** This is because I think the left right issue is related to the enemies' speed being a certain amount greater than the tile width (the enemy thinks left is shortest so it moves left, but it moves so far left than now right is the shortest option so it moves right, but...)

BryanGraeser:
I've doubled the frequency with which the enemy AI's movement is calculated. This should also help prevent the left-right issue (at the potential cost of performance). Performance is fine with one enemy, but may deteriorate with more.

lsizemore8:
Less an issue and more of a consideration. Right now, it's impossible to know when the enemy will stop chasing. It may be worth making the "glow" of the enemy decrease as the time before they stop moving decreases (so if the enemy is still in the player's light then the countdown will stay at the maximum value).

BryanGraeser:
> 
> 
> Less an issue and more of a consideration. Right now, it's impossible to know when the enemy will stop chasing. It may be worth making the "glow" of the enemy decrease as the time before they stop moving decreases (so if the enemy is still in the player's light then the countdown will stay at the maximum value).

Yeah I noticed this too and I don't have an easy answer. Definitely a UX question so we should ask Jalil. I was thinking maybe a marker of the location the enemy is investigating (ie the player's last known location a lot of the time).

foxhatleo (approved):
Looks good to me.

ronnie-dumesh (approved):
Approved

PR #47, opened 2020-03-27
Fix extra stepping

BryanGraeser:
LevelController was accidentally stepping world physics twice as much as it should be. Corrected this mistake.


lsizemore8 (approved):
Good catch

PR #49, opened 2020-03-28
Flare model

stallop:
Fixed flare collisions using preSolve and implemented a limit to the number of flares on screen at one time, changeable in the player json.


BryanGraeser (approved):
Max lmk what you think, but I think this sensor way is best

PR #50, opened 2020-03-29
Fps

lsizemore8:
Added support for a new "debug font" and am printing the fps count on the current exit position. 

ISSUE: Cannot get FPS in upper corner of screen because of the weird canvas coordinates. It is working and this is for internal purposes only. but I'm not fully sure how to make it perfect


BryanGraeser (approved):
I changed the printed value to be fps (1/delta) and decreased the frequency with which it is updated to make it more readable (it changing with every draw was hard to read).

I think it's gtg now

Issue #51, opened 2020-03-30
FlareCount issue with enemy model


BryanGraeser:
The initialize of flarecount: `setFlareCount(json.get("flarecount").asInt());` is currently handled in CharacterModel, which both PlayerModel and EnemyModel extend, so this causes a null pointer error when enemies are introduced because enemies do not have flares.

PR #52, opened 2020-03-30
move flare count to player model

BryanGraeser:
Solves issue #51.


stallop (approved):
Yeah this definitely makes more sense, good change.

PR #53, opened 2020-03-30
sneak and sprint working

stallop:
Got them both working as specified, implementation is a bit weird though. For example in movement, can't exactly just divide force by half if you want to sneak since that assumes you're currently in the walking state and wouldn't work properly if you were currently sprinting. This resulted in having to give the player forceSneak, forceWalk, and forceSprint variables and then updating their actual force variable with one depending on what movement state they're in.
Also had to create a variable to hold whatever the arbitrary sprint light radius value is (which I put in the player json), as well as a variable to hold the old radius from before the player starts sprinting so they will switch back to the same old light radius after they finish sprinting.
This is just how I imagined it would work, if there are any suggestions please let me know and I can make changes.

### DIFF
--- a/core/src/com/fallenflame/game/GameEngine.java
+++ b/core/src/com/fallenflame/game/GameEngine.java
@@ -300,6 +300,25 @@ public class GameEngine implements Screen {
             // Convert to radians with up as 0
             angle = (float)Math.PI*(angle-90.0f)/180.0f;
         }
+        if (level.getPlayer().getLightRadius() > 0) {
+            if (level.getPlayer().getForce() == level.getPlayer().getForceSneak()) {
+                // If player just stopped sneaking, set to walk
+                level.makeWalk();
+            }
+            if (input.didStartSprint()) {
+                // If player just started sprinting, store current light radius in lightRadiusNotSprint and change light radius to lightRadiusSprint
+                level.getPlayer().setLightRadiusNotSprint(level.getPlayer().getLightRadius());
+                level.getPlayer().setLightRadius(level.getPlayer().getLightRadiusSprint());
+                level.makeSprint();
+            } else if (input.didEndSprint()) {
+                // If player just stopped sprinting, set to walk and restore light radius to what it was before sprinting, which is in lightRadiusNotSprint
+                level.getPlayer().setLightRadius(level.getPlayer().getLightRadiusNotSprint());
+                level.makeWalk();
+            }
+        } else {
+            // If player light radius is 0, just sneak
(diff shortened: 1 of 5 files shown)


core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -300,6 +300,25 @@ public void update(float delta) {
             // Convert to radians with up as 0
             angle = (float)Math.PI*(angle-90.0f)/180.0f;
         }
+        if (level.getPlayer().getLightRadius() > 0) {
```
BryanGraeser: Can you refactor the `> 0` here to `> level.getPlayer().getMinLightRadius()` (I just added this function in PR)? The reasoning for this, is we want a minimum light radius that is greater than 0.

core/src/com/fallenflame/game/PlayerModel.java
```diff
@@ -11,10 +11,19 @@
 public class PlayerModel extends CharacterModel {
     /** Radius of player's light */
     protected float lightRadius = 0;
+    protected float forceSneak = 0;
+    protected float forceSprint = 0;
+    protected float forceWalk = 0;
```
BryanGraeser: I don't think we need to save these and I don't think we need setters for these. We can just have getters like so:
```    /**
     * Gets player force for sneaking
     * @return player force for sneaking
     */
    public float getForceSneak() {
        return getForce() / 2;
    }

**Review by BryanGraeser — CHANGES_REQUESTED** [review]
I think this implementation makes sense. Couple minor comments below that I'd like you to address

File: core/src/com/fallenflame/game/PlayerModel.java (hunk 2 above)
Comment (BryanGraeser) [reply core/src/com/fallenflame/game/PlayerModel.java]: Also lightRadiusSprint doesn't need a setter either. It can just be saved directly to the variable during initialization

File: core/src/com/fallenflame/game/PlayerModel.java [hunk 3]
```diff
@@ -11,10 +11,19 @@
...
+    protected float lightRadiusNotSprint = 0;
+    protected float lightRadiusSprint = 0;
 
     public void initialize(JsonValue json) {
         super.initialize(json);
         setFlareCount(json.get("flarecount").asInt());
+        setForceSprint(getForce() * 2);
+        setForceSneak(getForce() / 2);
+        setForceWalk(getForce());
```
BryanGraeser: So you can delete these

BryanGraeser (approved):
gtg

PR #54, opened 2020-03-31
JSON Standardization

BryanGraeser:
This PR adapts our json files and json file handling to reflect the data serialization now described on our architecture doc: https://docs.google.com/document/d/16eRPdaEX9A8tzajSBrh408LHoDjVKKBYueRL1Z9o_WA/edit# .

Please let me know what you think of this data serialization model, particularly if there is anything in the level json that should be in the global json or vice versa.

### DIFF
--- a/core/assets/jsons/global.json
+++ b/core/assets/jsons/global.json
@@ -1,0 +1,66 @@
+{
+  "player":  {
+    "radius":      	0.4,
+    "bodytype":    	"dynamic",
+    "density":        1.0,
+    "friction":       0.0,
+    "restitution":    0.0,
+    "force": 		 80.0,
+    "damping": 		 10.0,
+    "maxspeed": 	  5.0,
+    "startframe":       0,
+    "walklimit":		4,
+    "flarecount":       5,
+    "texture":     	"player",
+    "debugcolor":  	"white",
+    "debugopacity":	192
+  },
+  "enemies": {
+    "typeA":  {
+      "radius": 0.4,
+      "bodytype": "dynamic",
+      "density": 1.0,
+      "friction": 0.0,
+      "restitution": 0.0,
+      "force": 80.0,
+      "damping": 10.0,
+      "maxspeed": 5.0,
+      "startframe": 0,
+      "walklimit": 4,
+      "texture": "enemy",
+      "debugcolor": "white",
+      "debugopacity": 192
+    }
+  },
+  "exit": {
+    "size":           [ 1.25, 1.25],
+    "bodytype":    	  "static",
+    "density":  	  0.0,
+    "friction":       0.0,
+    "restitution":    0.0,
+    "texture":       "goal",
+    "debugcolor":    "yellow",
+    "debugopacity":   200
+  },
+  "flare" : {
+    "radius":      	  0.1,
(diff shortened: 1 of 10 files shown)


core/assets/jsons/global.json
```diff
@@ -0,0 +1,68 @@
...
+    "radius":      	0.4,
+    "bodytype":    	"dynamic",
+    "density":        1.0,
+    "friction":       0.0,
+    "restitution":    0.0,
+    "force": 		 80.0,
+    "damping": 		 10.0,
+    "maxspeed": 	  5.0,
+    "startframe":       0,
+    "walklimit":		4,
```
lsizemore8, line 12: What is the walk limit?

core/assets/jsons/global.json
```diff
@@ -0,0 +1,68 @@
...
+    "bodytype":    	  "dynamic",
+    "density":        1.0,
+    "friction":       0.0,
+    "restitution":    0.0,
+    "texture":        "fire",
+    "initialforce":   3.0,
+    "damping": 		 0.0,
+    "startframe":       0
+  },
+  "wall": {
```
lsizemore8, line 58: Why are walls both in the global json and the level json/why is the information overlapping between the two? Is it the global one is a default wall and the level one is for specialization only?

lsizemore8 (changes requested):
Left a couple questions. Main action item is the fact that walls are in both the global and local json, which I don't think should occur.

lsizemore8:
Also -> do we need to consider floors? This may be a design question, but unless we want the floor to be one color and we want no textures, we'll need a floor.

core/assets/jsons/global.json
BryanGraeser: There isn't any overlapping information except texture. I put all information that is constant for walls in the global json, and put all dynamic data (size, pos, and texture) in the level json. 

I did put texture also in the global json thinking that could be the default texture if one is not specified in the level json, but I'm not committed to that and may room it (also the code doesn't handle that case yet.

core/assets/jsons/global.json
BryanGraeser: This is the cooldown limit between walk animations

BryanGraeser:
Responded to your comments. Walls are meant to be in both, just like how player and enemy are in both (global constants vs dynamic data). 

For floors, that's a good point, but I think we can wait until designs are done so we have a better idea how that will be implemented.

core/assets/jsons/global.json
BryanGraeser: The default texture is now handled in WallModel (ie uses default if other is not specified).

lsizemore8: The "pad" attribute is overlapping in some walls (the second half of the walls in the level json), and it is the same value from the global. That's what I also meant in the "overlapping" aspect (in addition to texture).

BryanGraeser: Oh whoops, yeah I can fix that

lsizemore8 (approved):
Looks good to me. I'll approve it, but if anyone hasn't looked at it, please do before we merge it.

stallop (approved):
Structure makes sense to me and I don't spot any issues.

PR #55, opened 2020-04-02
Couple Architecture Changes

BryanGraeser:
1. Renamed LightInputProcessor to EventInputProcessor to better reflect it's role as the processor for event-based events (primarily mouse-wheel scrolling which changes the player's light radius)

2. Removed a cycle where EnemyModel had to access an enum defined in AIController. Now AIController unidirectionally depends on EnemyModel, not also the other way around.


lsizemore8 (approved):
Looks good.

PR #56, opened 2020-04-02
Abstract Enemies

BryanGraeser:
Abstracted AIController and EnemyModel. Any AI Controller will now extend abstract class AIController (for example: AITypeAController), and any Enemy Model will now extend abstract class EnemyModel (for example: EnemyTypeAModel). This is to allow us to add new enemy types going forward and edit existing enemy types without impacting others.

Further I switched action (currently just movement) representation to control codes from enums to allow movement & other actions (such as shoot) simultaneously later.


BryanGraeser:
While I was waiting I added flare-handling behavior to enemy TypeA, so that is now part of this PR as well.

ronnie-dumesh (approved):
Reviewed and approved

PR #57, opened 2020-04-03
Flare Fix

BryanGraeser:
Fixed flare json issue and also moved a couple of flare fields to global json now that we are holding all parameter data there.


stallop (approved):
All looks good!

PR #58, opened 2020-04-04
Diagonal movement 2

ronnie-dumesh:
This PR does two things.

The first is that it adds diagonal movement to the enemy's pathfinding. If enemies move in a diagonal direction, they move at a velocity of (+- .707) (as to maintain speed of 1)

The second is that it causes enemies to be viewed as red when chasing the player and green when investigating a player or last known position. This is done by changing the ILightRadius interface to ILight and to add support for a method getLightColor(). boolean EnemyModel.activated is now ActivationStates (a private enum) EnemyModel.state. The three ActivationStates are Calm, Alert, and Activated. These states do not necessarily correspond to the FSM of EnemyModel.

### DIFF
--- a/core/src/com/fallenflame/game/FlareModel.java
+++ b/core/src/com/fallenflame/game/FlareModel.java
@@ -38,6 +38,9 @@ public class FlareModel extends WheelObstacle implements ILight {
     /** Whether or not flare has stuck to wall */
     private boolean isStuck;
 
+    /**The color to tint the flare */
+    private Color tint;
+
     /**
      * Returns the light radius of this flare.
      *
@@ -50,7 +53,7 @@ public class FlareModel extends WheelObstacle implements ILight {
     /**
      * @return the color of the Flare's tint
      */
-    public Color getLightColor() {return new Color(255, 255, 255, 0.0f);}
+    public Color getLightColor() {return tint;}
 
     /**
      * Returns the directional movement of this flare.
@@ -197,6 +200,9 @@ public class FlareModel extends WheelObstacle implements ILight {
 //        int opacity = json.get("debugopacity").asInt();
 //        debugColor.mul(opacity/255.0f);
 //        setDebugColor(debugColor);
+
+        float[] tintValues = json.get("tint").asFloatArray();//RGBA
+        tint = new Color(tintValues[0], tintValues[1], tintValues[2], tintValues[3]);
 
         // Get the texture from the AssetManager singleton
         String key = json.get("texture").asString();
--- a/core/src/com/fallenflame/game/FlareModel.java
+++ b/core/src/com/fallenflame/game/FlareModel.java
@@ -10,7 +10,8 @@ import com.fallenflame.game.physics.obstacle.WheelObstacle;
 import com.fallenflame.game.util.*;
 import com.badlogic.gdx.graphics.*;
 
-public class FlareModel extends WheelObstacle implements ILightRadius {
+
+public class FlareModel extends WheelObstacle implements ILight {
     // Physics constants
     /** The force with which flare is originally thrown */
     private float initialForce;
@@ -45,6 +46,11 @@ public class FlareModel extends WheelObstacle implements ILightRadius {
     public float getLightRadius() {
         return lightRadius;
     }
+
+    /**
+     * @return the color of the Flare's tint
+     */
+    public Color getLightColor() {return new Color(255, 255, 255, 0.0f);}
 
     /**
      * Returns the directional movement of this flare.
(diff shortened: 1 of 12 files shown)


ronnie-dumesh:
I just noticed Leo's branch titled diagonal-movement. I believe this branch may have been what broke my build as that was when I was having all the github issues but I'm not entirely sure. In any case, I believe that branch could be deleted with Leo's approval as this implements all necessary diagonal movement functionality needed by the enemy

core/src/com/fallenflame/game/FlareModel.java
```diff
@@ -46,6 +47,11 @@ public float getLightRadius() {
         return lightRadius;
     }
 
+    /**
+     * @return the color of the Flare's tint
+     */
+    public Color getLightColor() {return new Color(255, 255, 255, 0.0f);}
```
BryanGraeser, line 53: Can you add this color as a parameter in the global json? That way we have an easy way to change it and the parameters are all consolidated.

BryanGraeser: Can you add this color as a parameter in the global json? That way we have all the parameters consolidated for easy adjustment.

core/src/com/fallenflame/game/PlayerModel.java
```diff
@@ -57,6 +58,14 @@ public float getLightRadius() {
         return lightRadius;
     }
 
+    /**
+     * Gets player color tint
+     * @return light color
+     */
+    public Color getLightColor() {
+        return new Color(255, 255, 255, 0.0f);
```
BryanGraeser: Can you add this color as a parameter in the global json too?

core/src/com/fallenflame/game/enemies/EnemyModel.java
```diff
@@ -18,31 +25,77 @@
...
-    /** Fire the ship weapon */
-    public static final int CONTROL_FIRE 	   = 0x10;
+    /** Move the ship to the down and left */
+    public static final int CONTROL_MOVE_DOWN_LEFT = 0x10;
+    /** Move the ship to the down and right */
+    public static final int CONTROL_MOVE_DOWN_RIGHT = 0x20;
+    /** Move the ship to the up and left */
+    public static final int CONTROL_MOVE_UP_LEFT = 0x40;
+    /** Move the ship to the up and right */
+    public static final int CONTROL_MOVE_UP_RIGHT = 0x80;
```
BryanGraeser: Couple questions here: Why have you chosen the control codes you did? 
And can you add back control_fire? I had that there because I am working on adding an enemy that shoots. (And we want the control_fire hex code to be able to combine with any movement code in case we have an enemy that moves and shoots.

Also can you change all the "ship" to "enemy" (I know that's layover from when I added it, but since you're changing this anyways

core/src/com/fallenflame/game/enemies/EnemyModel.java
```diff
@@ -18,31 +25,77 @@
...
      */
     public float getLightRadius() {
-        return getActivated() ? 1 : 0;
+        return isActivated() ? 1.0f : 0.0f;
+    }
+
+    public com.badlogic.gdx.graphics.Color getLightColor() {
+        if(isAgressive()){return com.badlogic.gdx.graphics.Color.RED;}
+        else if (isAlert()){return com.badlogic.gdx.graphics.Color.GREEN;}
+        else {return com.badlogic.gdx.graphics.Color.WHITE;}
```
BryanGraeser: Can you add these colors as a parameter in the global json?

BryanGraeser (changes requested):
Couple minor things

BryanGraeser:
> 
> 
> I just noticed Leo's branch titled diagonal-movement. I believe this branch may have been what broke my build as that was when I was having all the github issues but I'm not entirely sure. In any case, I believe that branch could be deleted with Leo's approval as this implements all necessary diagonal movement functionality needed by the enemy

Yeah, you should always be careful about using already existing branches, and if you do use one, make sure to reset it to the latest version of master.

core/src/com/fallenflame/game/enemies/EnemyModel.java
ronnie-dumesh: Control codes must be double of each other as it essentially does match statements by doing binary and operator. These are essentially the powers of 2 as hex codes.

ronnie-dumesh:
> Yeah, you should always be careful about using already existing branches, and if you do use one, make sure to reset it to the latest version of master.

Yeah, I actually didn't know that branch existed when I tried to make a branch diagonal-movement, which is where the issue came from. Will be more mindful next time

ronnie-dumesh:
Made all the necessary changes as for making data-driven colors and adding back shooting control code. Should be ready for approval. Master does not need to be merged.

BryanGraeser (approved):
gtg

PR #62, opened 2020-04-07
Architecture Modifcations - Input Controllers

BryanGraeser:
Updates to clean up architecture. 

Changes made:
- InputController and EventInputProcessor are now both folded into GameEngine so GameEngine directly handles all input instead of splitting it across 3 modules.
- Add support for comma and period changing of light radius


core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -445,5 +453,154 @@ public void setScreenListener(ScreenListener listener) {
...
+            horizontal += 1.0f;
+        }
+        if (Gdx.input.isKeyPressed(Input.Keys.A)) {
+            horizontal -= 1.0f;
+        }
+
+        // TODO: REMOVE CODE BELOW WHEN MOUSE WHEEL IS FIXED --> We should still include this as a functional alternative if possible
+        //#region mouse wheel alternative
+        if(Gdx.input.isKeyPressed(Input.Keys.PERIOD)){
+            lightRadius += 1.0f;
```
lsizemore8: This change is something that needed to be done anyways. I think this should changed to level.lightFromPlayer(1.0f), since right now this doesn't do anything because we don't use lightRadius at all.

core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -445,5 +453,154 @@ public void setScreenListener(ScreenListener listener) {
...
+        }
+        if(Gdx.input.isKeyPressed(Input.Keys.COMMA)){
+            lightRadius += -1.0f;
```
lsizemore8: See above, but same comment.

core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -30,63 +36,78 @@
...
+    private boolean sneakPressed;
+    private boolean sneakPrevious;
+    /** How much did we move horizontally? */
+    private float horizontal;
+    /** How much did we move vertically? */
+    private float vertical;
+    /** Did we increase or decrease the radius?
+     * Can only be two values: -1 and 1
+     */
+    private float lightRadius;
```
lsizemore8: If comments below are followed, this can be removed.

lsizemore8 (changes requested):
See comments. Now that we have everything in GameCanvas, we don't need lightRadius anymore and can just call level.lightFromPlayer(value)

BryanGraeser:
> 
> 
> See comments. Now that we have everything in GameCanvas, we don't need lightRadius anymore and can just call level.lightFromPlayer(value)

Made these changes. Changed the value to 0.5f for period and comma though because it is a lot less precise and gradual than the scroll wheel.

lsizemore8 (commented):
Looks good.

lsizemore8 (approved):
Yay for a large GameEngine

PR #63, opened 2020-04-08
Sound

stallop:
Directional sound kind of implemented w/ volume and pan, but either pan has hardly any effect or my headphones suck. Anyone feel free to check it out and let me know what you think and how you believe it could be improved.


BryanGraeser:
Where is SoundController used? It looks like it's not used at all, so I'm confused as to why it's included

BryanGraeser (approved):
Removed the extra SoundController file and I line I forgot to remove. I think it's gtg codewise now! Next if you can find free or temporary sounds that are closer to what they should be (and mono channel) would be great, but that can be a dif PR.

PR #64, opened 2020-04-09
Updated textures

ronnie-dumesh:
Updated filmstrips for the player's movement animations as well as setting the texture to be drawn offset from the physics body. Also added support for two different wall objects "vertical-wall" which uses the "single-brick" texture and "horizontal-wall" which uses the "double-brick" texture. These textures are being updated by Ravina. Also added support for the floor texture.


core/assets/jsons/global.json
```diff
@@ -33,17 +41,27 @@
...
       "startframe": 0,
       "walklimit": 4,
-      "texture": "enemy",
+      "texture":     	{
+        "left": "enemy",
+        "right": "enemy",
+        "up": "enemy",
+        "down": "enemy",
+        "walk-left": "enemy",
+        "walk-right": "enemy"
```
BryanGraeser: What is the difference between "left" and "walk-left" and between "right" and "walk-right"?

core/assets/jsons/global.json
```diff
@@ -72,14 +90,27 @@
     "tint": [1.0, 1.0, 1.0, 1.0],
     "startframe":       0
   },
-  "wall": {
+  "vertical-wall": {
```
BryanGraeser: Why the name switch between "vertical-wall" and "single-wall"? Why not just one or the other?

core/src/com/fallenflame/game/LevelController.java
```diff
@@ -590,7 +596,9 @@ public void draw(GameCanvas canvas, float delta, BitmapFont displayFont) {
 
         // Draw all objects
         canvas.begin();
-        player.draw(canvas);
+        //TODO: draw background here
+        //canvas.draw(,0,0)
+        levelModel.draw(canvas);
```
BryanGraeser: Why are we using level model to draw?

core/src/com/fallenflame/game/LightController.java
```diff
@@ -211,6 +211,18 @@ public void updateLights(Collection<FlareModel> flares, Collection<EnemyModel> e
...
+        //playerLight.setPosition(player.getX() + 300f, player.getY() + 300f);
+        //Changes being made when checking using get position?
+        //System.out.println(playerLight.getPosition());
+
+        //float xOffset = l.getX() + 0.65f *  (float) Math.abs(Math.sin(o.getAngle()));
+//        System.out.println(o.getAngle());
+//        System.out.println(l.getY());
+//        float yOffset = l.getY() + 0.65f * (float) Math.abs(Math.cos(o.getAngle()));
+//        System.out.println(yOffset);
+//        System.out.println("");
```
BryanGraeser: Please delete this

BryanGraeser (changes requested):
Couple minor comments, but my main question is why are we using LevelModel to draw?

core/assets/jsons/global.json
ronnie-dumesh: Was a difference originally, but deprecated now. Will remove it.

core/src/com/fallenflame/game/LevelController.java
ronnie-dumesh: Fixed

core/src/com/fallenflame/game/LightController.java
ronnie-dumesh: Done

core/assets/jsons/global.json
ronnie-dumesh: Good point, will fix it

core/src/com/fallenflame/game/LightController.java
ronnie-dumesh: Left the TODO for later however

BryanGraeser (approved):
Removed unnecessary level model changes to prevent later merge conflict. GTG now

PR #67, opened 2020-04-12
Fog

lsizemore8:
Fog is complete with minimal performance impacts. Every once in a while the fps does drop for a second, but that is likely due to the garbage collector and this problem is similar to the issue the professor mentioned in lab 3 for programmers. There is more fog around an enemy and no fog in the light radius/player position/walls position. 

It still is having a slight issue where its radius where it doesn't draw fog is larger than the actual light radius, but it works well enough.


BryanGraeser:
Okay, I haven't reviewed the actual code yet, but just from running it there are a couple things I noticed:
1. Can we slow down the movement of the fog a lot? I think it's quite distracting.
2. I can't tell that there's more fog around the enemy. I only notice it if I know exactly where the enemy is and look closely, expecting it to be there. There either needs to be a lot more, or it needs to be darker or something. The area around the enemy in which it exists also needs to be larger and randomized (ie so the enemy is not always perfectly in the center).
3. Fog is still around flares. We probably want it so that this is not the case.

But the fog avoiding walls and player light radius looks great and performance seems good.

core/src/com/fallenflame/game/GameCanvas.java
```diff
@@ -863,6 +864,30 @@ public void draw(PolygonRegion region, Color tint, float ox, float oy, Affine2 a
         computeVertices(local,region.getVertices());
     }
 
+//    /**@deprecated */
+//
+//    public void drawFog(ParticleEffectPool fogPool, FogController.fogParticle[][] fog, float delta){
+//        drawFog(fog, delta);
+//    }
```
BryanGraeser: Delete

core/src/com/fallenflame/game/LevelController.java
```diff
@@ -56,9 +61,10 @@
 
     // Controllers
     /** Light Controller */
-    private LightController lightController;
+    private final LightController lightController;
```
BryanGraeser, line 89: Why this change?

core/src/com/fallenflame/game/LevelController.java
```diff
@@ -56,9 +61,10 @@
...
     /** AI Controllers */
-    private List<AIController> AIControllers;
+    private final List<AIController> AIControllers;
+    private final FogController fogController;
```
BryanGraeser, line 92: Same question

core/src/com/fallenflame/game/LevelModel.java
```diff
@@ -172,9 +197,25 @@ public boolean inBounds(int x, int y) {
     public boolean isSafe(int x, int y) {
         //TODO: should return inBounds(x, y) and whether Tile.safe is true.
         //TODO: Tile.safe is false when it's an obstacle on it and true otherwise.
-        return (inBounds(x,y) && tileGrid[x][y].safe) || tileGrid[x][y].goal;
+        return (inBounds(x,y) && !(
+                        tileGrid[x][y].wall
+//                        tileGrid[x][y].enemy ||
+//                        tileGrid[x][y].player
+                )) || tileGrid[x][y].goal;
     } //TODO: temporary change
```
BryanGraeser: What are the comments here for? IsSafe is used by AI, so it should be safe even if a player or the enemy calling it is there

core/src/com/fallenflame/game/LevelController.java
```diff
@@ -466,8 +476,14 @@ public void update(float dt) {
                 }
             }
 
+            // Update level model.
+            levelModel.update(player, enemies);
+
```
BryanGraeser, line 554: This is fine, but please update earlier comments in this method that state "(and update levelModel)"

core/src/com/fallenflame/game/LightController.java
```diff
@@ -89,7 +89,7 @@ public void initialize(PlayerModel player, JsonValue levelLighting, World world,
         RayHandler.useDiffuseLight(true);
         rayhandler = new RayHandler(world, Gdx.graphics.getWidth(), Gdx.graphics.getHeight());
         rayhandler.setCombinedMatrix(raycamera);
-        rayhandler.setAmbientLight(0, 0, 0, 0);
+        rayhandler.setAmbientLight(0, 0, 0, 0.2f);
```
BryanGraeser: Try to set all constants like this as constants defined at the top of the class so they can be easily found and adjusted.

BryanGraeser (changes requested):
Some minor comments. We may have to wait on the changes I outlined in my earlier comment until after the presentation

core/src/com/fallenflame/game/LevelController.java
lsizemore8: @foxhatleo

foxhatleo: IntelliJ suggested and highlighted it with yellow background so I might have just clicked on the suggestion. I can revert it, altho it doesn't matter cos we don't create controller instance twice.

core/src/com/fallenflame/game/LevelController.java
foxhatleo: Same answer as above

foxhatleo (approved):
Good to go.

BryanGraeser (approved):
gtg

PR #70, opened 2020-04-13
Enemy typeb

BryanGraeser:
Added enemy typeb. This is the firing enemy. It has 3 states: Idle, Direct Fire (when firing at visible player), and Sustained Fire (when firing at player's last known position). Fireballs kill the player, generate light, and dissipate on collision with a wall. Refer to code comments for details, but major changes:
- Added EnemyTypeBModel
- Added AITypeBController
- Modified what is in EnemyModel vs extending classes
- Added control code for firing
- Added enemy firing handling in LevelController
- Added fireballModel 
- Added fireball handling to LevelController (including loss of game collision) and LightController


stallop:
The fireball speed seems to scale with the distance between the location fired at and the location of the enemy, is this intentional?

BryanGraeser:
> 
> 
> The fireball speed seems to scale with the distance between the location fired at and the location of the enemy, is this intentional?

Good catch. Should be fixed now.

stallop (approved):
That was the only issue I noticed, now that it's fixed everything looks good!

PR #72, opened 2020-04-13
Fixed Win Bug

### LINKED ISSUE #66: Player can lose if enemy catches up after winning
If the player wins and then the enemy catches up to the player, the game switches to "you died" text.

lsizemore8:
Fixed bug that caused the "victory" text to turn into "you died" if an enemy ran into you after you won.


core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -364,12 +364,12 @@ public void draw(float delta) {
...
 
         // Final message
-        if (isSuccess) {
+        if (!isFailed && isSuccess) {
             displayFont.setColor(Color.YELLOW);
             canvas.begin(); // DO NOT SCALE
             canvas.drawText("VICTORY!", displayFont, 0, canvas.getHeight());
             canvas.end();
-        } else if (isFailed) {
+        } else if (!isSuccess && isFailed) {
```
BryanGraeser: Doesn't this just mean nothing is displayed if the enemy runs into the player after winning?

BryanGraeser (commented):
Should we still allow anything to move after winning? Why don't we just stop the player from moving and levelController from updating?

BryanGraeser:
Changed it so the level freezes while the victory/loss message is displayed

BryanGraeser (approved):
gtg

core/src/com/fallenflame/game/GameEngine.java
lsizemore8: Yep. Tbh this pull request was done last night in the hopes that it would be merged in before today, it wasn't a permanent fix.

PR #74, opened 2020-04-13
Temp level selection and bug fixes

foxhatleo:
This is the branch of last-minute fixes of various things right before the presentation.
- Enemy type B uses our old player texture for now.
- Add temporary level selection using number keys: 1 for original, 2 for Jalil medium. Can plug any other levels in.
- Make "You die" and "Victory" text in the centre of the screen rather than moving around with the camera.
- Fix critical bug in LevelModel that causes crashes.


BryanGraeser (approved):
gtg

PR #75, opened 2020-04-13
Up/down directional sound really looks like it's not gonna work but h…

stallop:
Here is up/down sound as proposed. It's weird sounding and doesn't even convey up/down directional sound very well so I'm not sure if we should keep it tbh. Let me know if you have any ideas for things I could do to make it better.


BryanGraeser:
I think for now we can proceed without it, but if we find through playtesting that we really need up/down sound we can revisit this.

PR #76, opened 2020-04-16
names fixed, constant sound switched to one that doesn't cause popping

### LINKED ISSUE #69: Audio Popping
There is a popping noise in the audio. Perhaps it is due to audio files starting and stopping?

### LINKED ISSUE #71: Minor naming issue
This is quite minor, but technically the enemy "movement" sounds are really just enemy "active" sounds given that the non-moving enemy plays these sounds when active as well.

stallop:
Changed move to active for naming. The audio popping appeared to be caused by the enemy constant sound being played along with others, replacing it with different sounds seems to fix the issue so it's just a dog growl for now until we start really deciding on sounds. Also converted everything to .wav format since that's best for a desktop game, and removed all the unnecessary unused sound files we were accumulating.


BryanGraeser (approved):
gtg

PR #77, opened 2020-04-16
Ghost Enemy

BryanGraeser:
Currently the ghost:
- enemy immediately chases the player
- is blocked by the player's light radius unless the player is sneaking. IE it can only kill you when you're sneaking **unless you walk into it (do we want it to be actively repelled by the light?)**
- moves at a slower speed than the other moving enemy

I temporarily added it to the medium level for testing, but will remove before merging

### DIFF
--- a/core/assets/jsons/global.json
+++ b/core/assets/jsons/global.json
@@ -94,6 +94,38 @@
         "y": 0
       },
       "tintOpacity": 192
+    },
+    "ghost":  {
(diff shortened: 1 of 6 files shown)


BryanGraeser:
So big question is whether we want the ghost to run away from light

BryanGraeser:
Behavior has been changed via team discussion

PR #79, opened 2020-04-18
changed getX() and getY() for CharacterModel objects to be center…

ronnie-dumesh:
light radius should now be at the center of the CharacterModel and lights should originate from center


BryanGraeser (changes requested):
Collisions are no longer behaving properly. Turn on debug mode with 'g' and run up or down into a wall and you'll see. Up it doesn't go close enough, down it goes too far. So the debug texture isn't being drawn correctly and the physics body is still at the center of the light

BryanGraeser:
In the future, try to test your change before making a PR. That way if it takes me a bit to review your PR, you don't have to wait until then to be aware of bugs

ronnie-dumesh:
> In the future, try to test your change before making a PR. That way if it takes me a bit to review your PR, you don't have to wait until then to be aware of bugs

I mean I do, I just wasn't sure what intended behavior was there. 

Let me know if you're happy with how much the player is touching walls that are below them. If you want to change it, just fiddle with line 29 of global.json (player.textureoffset.y)

ronnie-dumesh:
I made some decently big changes to lightController. They could potentially impact performance but I still had ~60 FPS when I tried to "stress test". I've tagged @foxhatleo so he can take a look if anything stands out as bad to him.

BryanGraeser (approved):
gtg

PR #80, opened 2020-04-21
Ghost enemy

BryanGraeser:
New behavior ready! So this ghost:
- Appears after the player has used up all of their sneak
- Chases the player down, moving directly towards them through walls and enemies
- Kills the player on contact

The sneak value is set in the global JSON and determines how long the player can sneak "for free". This is not yet displayed anywhere bc we don't have an official design for that yet. The player is also still able to sneak after the ghost is triggered (and sidenote the sneak value will keep decrementing to negative values), so the sneak value currently only exists to trigger the ghost. I can make the sneakvalue reaching 0 also prevent further sneak if we wish.

One question: should the sneak value be tied to level or player (global)? Rn it's global


BryanGraeser:
> One question: should the sneak value be tied to level or player (global)? Rn it's global

Sneak value is now taken from level JSON, but if no value is specified in level JSON, it defaults to a value in the global JSON. This is to ensure backwards compatibility with already generated levels

core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -373,7 +373,9 @@ public void update(float delta) {
         }
         if (sneakPressed && !sneakPrevious) {
             // If player just started sneaking
-            level.makeSneak();
+            sneakPressed = level.makeSneak();
+            // Sets sneakPressed to makeSneak() return fo sneakPressed is false if player could not sneak
```
lsizemore8: I believe we said the player can still sneak even if they are "out", but instead the ghost would just come

core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -373,7 +373,9 @@ public void update(float delta) {
         }
         if (sneakPressed && !sneakPrevious) {
             // If player just started sneaking
-            level.makeSneak();
+            sneakPressed = level.makeSneak();
```
lsizemore8: Update: does not actually work, I can still sneak when ghost is here

lsizemore8 (changes requested):
Update the comments in the code to state whether we are allowing sneak once ghost arrives or not. I get why you did it this way, it's a quick change to go from infinite sneak, but add a comment or two stating that it currently allows infinite sneak.

core/src/com/fallenflame/game/GameEngine.java
BryanGraeser: Oh whoops, that's left over code. I will remove it (but the behavior now is as you describe)

core/src/com/fallenflame/game/GameEngine.java
BryanGraeser: Yeah, I changed it so you can still sneak, but forgot to remove some of this stuff

BryanGraeser:
Should be removed now! Behavior is the same, but that code was now superfluous so good catch

ronnie-dumesh (approved):
Reviewed fix, looks good

lsizemore8 (approved):
Lit lit

PR #81, opened 2020-04-23
Resizable bounds

foxhatleo:
This branch implements resizable bounds. It also makes changes to LevelController and FogController for optimisation. Some important changes:

- Draw scale is now always 50x50. 
- "graphicSize" is now removed from the level JSON. We didn't use it anyways.
- Some minor changes to light controller but not much effects on the result.
- Fog controller has some new optimisation code added.
- **IMPORTANT!** Fog controller's updateFog() and draw() are combined into "updateFogAndDraw()". This is because during the update phase the camera of the canvas is now necessary because of optimisation. Because level controller doesn't have the canvas unless in "draw()", it is only possible to update the fogs during draw phase too. Let me know how you lot feels about this.

With the optimisation tricks, the game can maintain ~60 fps for lower-end triple digits map (I changed easy level from [16,12] to [160,120] to test), but it will slow down at [1600,1200], which I don't think we need. I don't see that this branch affects any behaviour such as how fog looks, etc. compared to master now, but it'd great if someone tests it out.


core/src/com/fallenflame/game/FogController.java
```diff
@@ -60,9 +82,10 @@ public void updateFog(Vector2 scale) {
...
             }
         }
+        draw(canvas, delta);
     }
 
-    public void draw(GameCanvas canvas, float delta) {
+    private void draw(GameCanvas canvas, float delta) {
         canvas.begin();
         canvas.drawFog(fog, delta);
         canvas.end();
```
BryanGraeser: Very very minor, but this probably doesn't need to be a separate function anymore

BryanGraeser (changes requested):
There are is a weird performance bug I'm noticing where fog doesn't show up in certain areas of the map.
When you first spawn everything looks fine:
image

But as you walk to the right you see an area that is missing a lot of fog:
image
image
(debug version so you can see walls)

Then if I walk up to the top and around, the start left area no longer has fog:
image

Something peculiar is happening with updating. It doesn't seem to be correctly updating based on the camera position. And the ambient, light fog is disappearing.

BryanGraeser:
Ignoring the above issue, the changes to fog controller look good and make sense to me.

foxhatleo:
I see the issue. This is most peculiar because the bound checking part seems to be causing the problem. The bound-checking optimisation is done by clearing all particles of a tile when it is outside the camera. Removing this part solves the issue but we don't get the necessary performance fix.

However, the conditional check logic seems fine. I tested this by printing true/false (whether a tile is outside a map or not), and the check seems to be working as expected. Therefore, it might be an issue in the particle pool thing, which I need to talk to @lsizemore8  about.

The part of code I am talking about in `updateFogAndDraw`:
```java
// If this tile is not in camera, clear its content.
if (x < lowX || x >= highX || y < lowY || y >= highY) {
    if (fog[x][y] != null) {
        for(ParticleEffectPool.PooledEffect effect: fog[x][y].fogParticles){
            effect.free();
        }
        fog[x][y].fogParticles.clear();
    }
    continue;
}
```

It is much too late here so I will work on this tomorrow. And Laura if you can help me with this that'd be great.

lsizemore8:
> I see the issue. This is most peculiar because the bound checking part seems to be causing the problem. The bound-checking optimisation is done by clearing all particles of a tile when it is outside the camera. Removing this part solves the issue but we don't get the necessary performance fix.
> 
> However, the conditional check logic seems fine. I tested this by printing true/false (whether a tile is outside a map or not), and the check seems to be working as expected. Therefore, it might be an issue in the particle pool thing, which I need to talk to @lsizemore8 about.
> 
> The part of code I am talking about in `updateFogAndDraw`:
> 
> ```java
> // If this tile is not in camera, clear its content.
> if (x < lowX || x >= highX || y < lowY || y >= highY) {
>     if (fog[x][y] != null) {
>         for(ParticleEffectPool.PooledEffect effect: fog[x][y].fogParticles){
>             effect.free();
>         }
>         fog[x][y].fogParticles.clear();
>     }
>     continue;
> }
> ```
> 
> It is much too late here so I will work on this tomorrow. And Laura if you can help me with this that'd be great.

Particle Pool doesn't really do anything (it creates a new object if there is too little in the pool). That shouldn't do anything. I can take a look as well to see what is going on

foxhatleo:
Laura's fog fixes seems to fix the issue and it looks just like master now from what I can tell.

lsizemore8 (approved):
Given Max just tested this for us with a resized level and said it works, I'm approving it. However, @BryanGraeser has to approve it before we can merge to master, since he requested changes

BryanGraeser (approved):
gtg. Haven't looked in-depth but given others reviewed and tested I'll just approve

PR #83, opened 2020-04-23
Pathing enemy

BryanGraeser:
Pathing enemy! When in an idle state, the enemy will follow a set of predefined points. When distracted by a flare or player, the enemy will give chase, and when finished investigating, will return to their pathing.

To make a pathing enemy, **in the level JSON** for the desired enemy set field `"enemytype"` as `"typeA"` and then set the field `"subtype"` to `"pathing"`. This will require you to then create and populate a `"pathCoors"` field with coordinates for the enemy, defined like this example: 
`"pathCoors": [
        [7.7, 5.1],
        [14.45, 1.34],
        [14.45, 7.61],
        [5, 5]
      ]`

**Note: this will require addition of new level editor support**, however this is backwards compatible beacuse if no `"subtype"` is defined, it will be interpreted as default `"typeA"` (aka our current typeA enemy.

### DIFF
--- a/core/assets/jsons/levels/easy.json
+++ b/core/assets/jsons/levels/easy.json
@@ -30,7 +30,14 @@
         5,
         5
       ],
-      "enemytype": "typeA"
+      "enemytype": "typeA",
+      "subtype": "pathing",
+      "pathCoors": [
+        [7.7, 5.1],
+        [14.45, 1.34],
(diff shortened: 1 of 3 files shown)


foxhatleo (approved):
Looks good to me.

This will take this weekend to implement in level editor btw.

PR #84, opened 2020-04-23
fix bug and move shared function to AIController

### LINKED ISSUE #82: Incorrect light detection for enemies
Enemies' AIControllers check if they are within a player's light radius, and if so, activate the enemy. We recently moved the light to center on the player's texture, however it seems enemies are still detecting as if the light was centered on the player's physics body.

For example, in this position the enemy "saw" the player, despite not being actually in the player's light radius:
image

I think this is a result of PR #79.

BryanGraeser:
Fixes #82. Issue was AIController was checking player coordinates which after PR #79, are no longer descriptive of player light center.

Also refactored helped function checking if enemy is within player light radius to AIController bc it is shared across mult AI types and to facilitate this I moved the playermodel handling to AIController.

This (having light centered on player) does raise a gameplay issue where if you are walking down, because the player is so close to the bottom of the light radius, you can die basically instantly.

### DIFF
--- a/core/src/com/fallenflame/game/enemies/AIController.java
+++ b/core/src/com/fallenflame/game/enemies/AIController.java
@@ -1,6 +1,7 @@
 package com.fallenflame.game.enemies;
 
 import com.fallenflame.game.LevelModel;
+import com.fallenflame.game.PlayerModel;
 
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


ronnie-dumesh (approved):
In withinPlayerLight(), you should use enemy.getTextureX() as potentially the enemies' textures will not be centered around the body as they are now. Looks good otherwise.

BryanGraeser:
Good point, will make that change

PR #85, opened 2020-04-23
Delete Gameplay Prototype

BryanGraeser:
I think it's time...

(also note you can still access this if you really need via old git commits)


ronnie-dumesh (approved):
Approved with great enthusiasm

PR #86, opened 2020-04-24
Level select mode

stallop:
Level select now works but there's a bug where the click from selecting a level is getting carried over into the engine screen and shooting a flare. Pretty sure I'm doing something wrong with inputProcessors as that part of our code kind of confused me.


core/src/com/fallenflame/game/GDXRoot.java
```diff
@@ -105,13 +109,21 @@ public void exitScreen(Screen screen, int exitCode) {
 		if (screen == loading) {
 			//Still finish loading everything before shutting down
 			engine.loadContent();
+			Gdx.input.setInputProcessor(levelSelect);
```
lsizemore8, line 112: I'm testing it now, but I don't think we need any of this, as that's the purpose of the multiplexer.

lsizemore8: Nevermind that makes no sense, but it did not work.

BryanGraeser (approved):
gtg

PR #87, opened 2020-04-24
Exit light

BryanGraeser:
Exit glows now! It's a very small glow and I also think it is fine with fog over it.

FYI will probably cause a merge conflict with #81, but the changes to lightController aren't too significant here.


stallop:
Looks good, especially with the circular exit texture. I think whenever we eventually replace the placeholder texture, it might be a good idea to keep that shape as the circular glow fits it really well.

BryanGraeser:
Good point on the circle texture

stallop (approved):
Looks good!

PR #88, opened 2020-04-25
Fog fixes

lsizemore8:
Removes fog from flare light and adds a wider radius of fog around enemies.


core/src/com/fallenflame/game/FogController.java
```diff
@@ -50,11 +69,15 @@ public void updateFog(Vector2 scale) {
...
-                    for (int i = 0; i < (1 + (levelModel.hasEnemy(x, y) ? 8 : 1)); i++) {
-                        float randomVal = (float) (Math.random() * TILE_SIZE);
-                        effect.setPosition((x * TILE_SIZE + randomVal) * scale.x, (y * TILE_SIZE + randomVal) * scale.y);
+                    float incX = (float) ((Math.random() - 0.5) * 2) * 3;
+                    float incY = (float) ((Math.random() - 0.5) * 2) * 3;
+                    for (int i = 0; i < (1 + (levelModel.hasEnemy(x, y) ? 11 : 1)); i++) {
+                        float randomVal = levelModel.hasEnemy(x,y) ? 2.0f : 1.0f;
+                        float randomX = (float) (((Math.random()*randomVal)-0.5f) * TILE_SIZE);
+                        float randomY = (float) (((Math.random()*randomVal)-0.5f) * TILE_SIZE);
+                        effect.setPosition(((x+incX) * TILE_SIZE + randomX) * scale.x, ((y+incY) * TILE_SIZE + randomY) * scale.y);
```
BryanGraeser: Can we factor out randomization constants here like the 0.5, 2, and 3 from `((Math.random() - 0.5) * 2) * 3` and have them defined as constants at the top of the class with descriptions on what their values change? That way we can more easily tweak the randomization to our likes. Also why is it *2 then *3? Why not *6?

BryanGraeser (changes requested):
Like I mentioned to Laura, I think we should get fog-speed merge in here first. Also we need to make it easier to adjust randomization constants

core/src/com/fallenflame/game/FogController.java
lsizemore8: Lmao what happens when tired coders try to do math. Resolved in new push

BryanGraeser (approved):
Looks gtg. Also added a very minor change removing an unrelated println I noticed.

PR #89, opened 2020-04-25
Shooting enemy flare behavior

BryanGraeser:
New behavior for shooting enemy:
1. Shooting enemy starts shooting at flare when the flare's light radius touches it
2. Continues shooting at flare (even if out of light radius) until the flare hits a wall 
3. Once the flare hits a wall, it fires one last shot

Additionally, the shooter prioritizes flares over the player as targets.


lsizemore8 (approved):
Looks good to me

PR #90, opened 2020-04-25
Make runner prioritize flares and change flare burnout behavior

BryanGraeser:
Running enemy now:
- Prioritizes flares over the player 
- Whenever it can see a flare, it will ignore the player in favor of it

Changed flare burnout behavior as well for balancing. Burnout countdown now does not start until flare has hit a wall. I also adjusted the timeout time and speed for balancing.


stallop (approved):
Everything here seems good.

PR #91, opened 2020-04-26
Fog speed

lsizemore8:
New fog texture implemented and removed fast "particle" like feeling. This needs to be merged in first before fog-speed and the bounds issue. 

Issue (will add to issue tracker): every 3000 seconds or so, the fog will all disappear and reappear. I suspect that has to do with the life of the particle, and sans periodically removing/adding a particle (and that may cause the bug from before), there is not much I can do given how the ParticleEffect is structured. Will play around with this if time.


BryanGraeser:
> 
> 
> New fog texture implemented and removed fast "particle" like feeling. This needs to be merged in first before fog-speed and the bounds issue.
> 
> Issue (will add to issue tracker): every 3000 seconds or so, the fog will all disappear and reappear. I suspect that has to do with the life of the particle, and sans periodically removing/adding a particle (and that may cause the bug from before), there is not much I can do given how the ParticleEffect is structured. Will play around with this if time.

Do you mean merged in before fog-fixes? And I agree we can worry about that later.

BryanGraeser (approved):
Looks good! Nice job :)

PR #94, opened 2020-04-27
Update enemy textures

ronnie-dumesh:
A couple of textures need to be updated, but that's with the designers. It's one-line fixes for me.

This code updates textures to enemy types A and B, updates the player flare texture, and adds the fire buddy texture to the player.

The commits from before 04/25 are from me forgetting to merge master before making a new branch


core/src/com/fallenflame/game/LevelController.java
```diff
@@ -526,6 +528,11 @@ public void update(float dt) {
                     else
                         ((EnemyTypeBModel)enemy).coolDown(true);
                 }
+                if(enemy.getClass() == EnemyTypeBModel.class){
+                    ((EnemyTypeBModel)enemy).update(dt);
+                } else {
+                    enemy.update(dt);
+                }
```
BryanGraeser: Why is this necessary? You're making the same function call from levelcontroller's perspective. You don't need to type cast as EnemyTypeBModel, if that's what it is and EnemyTypeBModel overrides update, it will be call the override version. 
You can replace this whole section with `enemy.update(dt);`

core/src/com/fallenflame/game/enemies/EnemyModel.java
```diff
@@ -83,6 +83,11 @@ public void initialize(JsonValue json, float[] pos) {
         }
     }
 
+    @Override
+    public void initializeTextures(JsonValue json){
+        super.initializeTextures(json);
+    }
+
```
BryanGraeser: Why? You're overriding it just to call the version you overwrote? You can just delete this

BryanGraeser (changes requested):
Couple small changes listed below.

Was hoping rebasing would remove some of the extraneous differences from earlier commits, but it didn't. So yeah, next time try to make a new branch up-to-date with master for changes so the PR doesn't have a bunch of changes listed that aren't really changes.

core/src/com/fallenflame/game/enemies/EnemyModel.java
ronnie-dumesh: I can't have an override method EnemyTypeBModel.intializeTextures() if the direct parent doesn't have one. So it's not enough for CharacterModel to have initializeTextures(), EnemyModel needs it as well.

ronnie-dumesh: Nvm, I tested it out rn and it worked. For some reason the IDE was throwing errors though. I'll remove this

core/src/com/fallenflame/game/LevelController.java
ronnie-dumesh: Fixed

BryanGraeser (approved):
gtg

PR #96, opened 2020-04-27
New sound effects and introducing background music

foxhatleo:
This PR adds new sound effects. Old sound effects are still there for possible compatibility issues, but global JSON has been updated to use the new sound effects.

Furthermore, background music for menu is added. Level-specific background music is also added. There are 12 pieces of background music in total to choose from at the moment (including the one for menu). Old JSONs for levels will still work because by default the level will play no music if the "bgm" attribute is not found.

All new music are located in "core/assets/sounds/new-sounds". A "license" file explains where the music or sound fx comes from. Everything is free-of-charge, royalty-free, modification and commercial allowed.

The addition of new assets does add more time to load the game initially, but otherwise no performance issue is observed.

This PR also removes SoundController from the repo cos it seems that we are not using it at all. Lmk if that controller is still needed, especially @mstallop


BryanGraeser (changes requested):
Couple things:
- The ghost enemy needs sounds, rn it's still the barking dog. It can prob be the same sounds as other enemies
- The music is too loud. Can you turn it down?

foxhatleo:
- Background music has been turned down.
- A new sound is added for ghost.
- Ghost now has a special "background music" attribute in the global JSON, where when the ghost is activated, level-specific music will stop and globally-defined ghost-activated music would play. It has a faster tempo to tell the player that death is coming. This is entirely optional and if a ghost bgm is not defined it will just ignore it and keep playing level music instead. Rn I have included a different music for the ghost.
- Some more bug fixes are done.

BryanGraeser (approved):
Nice!

PR #97, opened 2020-04-28
Temp Solution to go back to menu after completing level

BryanGraeser:
This is a temporary solution for the presentation tomorrow. You can now press `esc` to exit the level and go back to level select.
Current issues:
- **Sounds keep playing.** For some reason, unable to stop enemy sounds, other models don't have a way to get active sounds. This needs to be fixed, but may not be an issue for the presentation.
- I had to create 2 GameCanvas's in gdxRoot: one for GameEngine and one for LevelSelectMode. This is because Engine does things to the GameCanvas that I am having trouble reverting (such as moving the camera and changing font colors etc). 

This solution is not the most ideal, but works for now.

### DIFF
--- a/core/src/com/fallenflame/game/GameEngine.java
+++ b/core/src/com/fallenflame/game/GameEngine.java
@@ -321,7 +321,7 @@ public class GameEngine implements Screen, InputProcessor {
             }
         }
         if (exitPressed && !exitPrevious) {
-            listener.exitScreen(this, EXIT_QUIT);
+            listener.exitScreen(this, 0);
             return false;
         }
         //If countdown is > -1, then the player must have won or lost. Either continue to show the win condition message
@@ -460,6 +460,7 @@ public class GameEngine implements Screen, InputProcessor {
      * also paused before it is destroyed.
      */
     public void pause() {
+        level.stopAllSounds();
        isPaused = true;
     }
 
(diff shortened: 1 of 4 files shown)


stallop (approved):
We can improve upon it later and it seems like it should work fine for the demo tomorrow.

core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -460,6 +460,7 @@ public void render(float delta) {
      * also paused before it is destroyed.
      */
     public void pause() {
+        level.stopAllSounds();
```
lsizemore8, line 463: This doesn't seem to actually do anything, as I still hear the dog sounds when I'm paused.

lsizemore8 (changes requested):
Dog sound still plays in background when paused. If that isn't easily fixed I can undo this since it won't matter for tomorrow's presentation, but if it's a simple fix then that's a bug.

lsizemore8 (approved):
Oops I'm blind, didn't see that the enemy sound was known. It's fine for tomorrow

core/src/com/fallenflame/game/GameEngine.java
BryanGraeser: lol yes this exactly what I wrote

PR #99, opened 2020-04-28
Add hud elements

ronnie-dumesh:
Adds an indicator for how much sneak a player has to the top-left and the an indicator for the number of flares a player has to the top-right. Both have maximum values (1000 and 5 respectively) that can be modified in global.json.


BryanGraeser (changes requested):
Flare indicators don't disappear as they are used

BryanGraeser (approved):
gtg

PR #100, opened 2020-04-28
new levels, issue with createFlare

stallop:
Issue with createFlare getting mouse click pos relative to window dimensions on large stages.  When calculating the center screen position it incorrectly uses bounds.x and bounds.y which are not actually the window dimensions, but rather the level dimensions. Once that is fixed these levels should be good to go other than cosmetic issues with textures.


BryanGraeser (changes requested):
Level 3 has no enemies

BryanGraeser (approved):
Whoops I'm dumb, it's supposed to be like that

PR #101, opened 2020-04-28
Speed up loading time

BryanGraeser:
Removed lots of sound assets being loaded. Many of them weren't used. I also changed all levels to use the menu-music so right now that is the only background music being loaded. In the future, we can add back other background sounds/music we want to load.


ronnie-dumesh (approved):
Works much faster

PR #102, opened 2020-04-28
fixed crashing upon ghost

ronnie-dumesh:
Crashing happened when ghost loaded in. This PR modifies global.json to prevent this


BryanGraeser (approved):
Whoops forgot the music tied to the ghost. Thanks!

PR #103, opened 2020-04-28
Updated enemy textures 2

ronnie-dumesh:
-Adds new sneakmeter textures to top-left of screen
-Adds new flarecounter textures to top-right of screen
-Fixes shooter textures from never reseting once triggered
-Has fire buddy come towards player while sneaking


core/src/com/fallenflame/game/PlayerModel.java
```diff
@@ -34,7 +34,7 @@
     protected float lightRadiusSaved;
     protected float lightRadiusSprint;
     protected float lightRadiusSneak;
-    /** Player sneak left (once hits 0, player cannot sneak) */
+    /** Player sneak left (once hits 0, a ghost is deployed on the map) Sneakval must be greater than 0 */
```
BryanGraeser: Must be greater than or equal to 0. It can hit 0 because that's when ghost appears

BryanGraeser (approved):
gtg

PR #104, opened 2020-04-29
Make flares a limited resource

BryanGraeser:
Made player have set number of flares at start of level that is decremented every time they fire. Currently no way to regain flares (working on pick-up).

Also please note for **level editor**:
- `"startFlareCount"` defines how many flares the player starts with
- `"maxFlareCount"` defines the max flares a player can hold and what the UI shows (will come in to play with pick-ups)
I differentiated these in case we want a level where you start out with few, but pick up more later. Also note, both of these have **global default values of 5 to allow for backwards compatibility.**

### DIFF
--- a/core/assets/jsons/global.json
+++ b/core/assets/jsons/global.json
@@ -10,7 +10,7 @@
     "maxspeed": 	  5.0,
     "startframe":       0,
     "walklimit":		4,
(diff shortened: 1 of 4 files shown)


lsizemore8 (approved):
Once master is merged, this should be good to go

PR #105, opened 2020-04-29
Ui fixes

lsizemore8:
Updated LoadingMode to implement most of UI for loading and menu selection. 
Some notes: 
1. The font colors are not completely correct. Given how libgdx works, to get the exact color (with RGB), I would have to get this text file with white text and a transparent background. There's supposedly a Libgdx executable (like the particle editor) that does this, but I am still looking for it. The colors that are in right now are close, but not exact. 
2. All menu options go to Level Select rn. I left that as is for now to get feedback, but likely I'll have to just make clicking "Controls" do nothing. 
3. The loading bar is not the designed loading bar for this, but the legacy code for the loading bar is wack, and I don't have much time to work on this again before Thursday. Will add it after Thursday.


BryanGraeser (approved):
Approved but a couple things to keep in mind:
- This does seem to slow down loading. Not sure if there's anyway to avoid this, but it also seems odd bc the only assets you added are loading mode stuff which is loaded first so why is it slowing down later loading?
- There is an issue #106 with font coloring, but it was not introduced by this PR

BryanGraeser:
If you can find anyway to speed up loading that'd be great

stallop (approved):
Slow loading time is strange and hopefully fixable, but visually everything seems good.

lsizemore8:
> Approved but a couple things to keep in mind:
> 
> * This does seem to slow down loading. Not sure if there's anyway to avoid this, but it also seems odd bc the only assets you added are loading mode stuff which is loaded first so why is it slowing down later loading?
> * There is an issue #106 with font coloring, but it was not introduced by this PR

I think this will have to be refactored into 2 files: LoadingMode and MenuSelectMode, since I suspect it's the adding of the BitmapFont that is causing the slowdown. I don't have time before Friday to get that done due to other classes, but I'll refactor that the second I can. I'm gonna merge this in for now but I think that's what caused it. That's honestly on me for poor design from the get go, was definitely trying to save some refactoring time.

PR #107, opened 2020-04-29
Level select color

### LINKED ISSUE #106: Text changes color when you finish level (win or lose)
The text starts out white, but the win/lose message is displayed in red. This causes the level select text to be red when you exit to it. 

A hacky fix would be to make win/lose message white as well, but probably better to address source of issue.

lsizemore8:
fixes #106  where level select mode was not resetting the color to white


stallop (approved):
Good change

PR #109, opened 2020-04-30
Floor & reset fix

foxhatleo:
This PR fixes two issues:

- Wall texture cuts off when level bounds get bigger than screen
- Press "R" resets to the first level rather than the current playing level


BryanGraeser (approved):
gtg

PR #110, opened 2020-04-30
ItemModel

BryanGraeser:
ItemModel is now implemented. For now it can only represent flares, but it is designed to work for future items as well.

Functionality for flares:
- If player walks on it while they are not at max flares, they get a new flare and the item disappears
- If player walks on it while they are at max flares, they do not get a new flare and the item remains (if they fire a flare while still standing on it, they will pick it up).

To add items to level editor use a field in the JSON structured like this:
"items": [
    {
      "itemPos": [3, 1.5],
      "itemType": "flare"
    },
    {
      "itemPos": [7, 5],
      "itemType": "flare"
    }
  ]
`itemPos `is the position and `itemType `can only be `"flare"` for now.


BryanGraeser:
Items also have a radius 1 glow to be more visible

PR #112, opened 2020-05-01
Minor texture update

ronnie-dumesh:
Updates several textures such as exit, the player left and right animations, enemy up, the sneakbar background, and the shooter animation.


lsizemore8 (approved):
Looks good!

PR #113, opened 2020-05-01
added enemy type B to levels, centered text on level select

stallop:


lsizemore8 (approved):
Woot woot this seems to check out

PR #114, opened 2020-05-01
New levels

lsizemore8:
Added the three new levels that we talked about in the meeting. I also added the "locked" image for the levels we didn't have yet, so it would be more obvious which were not available for tomorrow

Note: the test levels are now levels 7 and 8.


BryanGraeser (approved):
didn't test, merging for release

PR #116, opened 2020-05-01
New Movement scheme and constants changing

BryanGraeser:
Movement scheme is now instantaneous velocity change rather than force acting over time.


lsizemore8 (approved):
Yay for so many fixes

PR #118, opened 2020-05-02
Add return to main menu, rebind control, etc

### LINKED ISSUE #98: Exit level to level select
Introduced by PR #97 

Current issues:
- Sounds keep playing. For some reason, unable to stop enemy sounds, other models don't have a way to get active sounds. This needs to be fixed, but may not be an issue for the presentation.
- I had to create 2 GameCanvas's in gdxRoot: one for GameEngine and one for LevelSelectMode. This is because Engine does things to the GameCanvas that I am having trouble reverting (such as moving the camera and changing font colors etc).

foxhatleo:
This PR:

- Adds back button to main screen.
- Adds a control screen.
- Allow users to rebind keyboard (the preferences of key will save even when game is closed)
- Make background black instead of blue
- Fix a floor rendering bug
- Add a pause screen

Resolves #98


BryanGraeser (commented):
This is a lot of excellent changes! Nice job. I just have 2 comments:
- Currently the control customizing does not support 'spacebar', I think we should support this if possible as users indicated this was a desired key binding
- I think we should describe the mouse controls on the control screen (so scroll wheel up/down and mouse click). I understand we likely want these to be uneditable, but I think they still need to be listed here as the controls screen is meant to communicate controls to the player as well as provide a way to edit them. Perhaps list them under the "click on a key to change it" text? I know this may not fit, but we are also increasing the screen size so perhaps we can wait until then to add this.

BryanGraeser (approved):
I'm going to go ahead and approve this because the above changes can be added later and don't affect the rest of this PR which has a lot of important changes.

PR #119, opened 2020-05-03
Add text support

foxhatleo:
This PR adds text support. Level editor will be updated soon to support this.


core/src/com/fallenflame/game/TextController.java
```diff
@@ -0,0 +1,87 @@
...
+import org.w3c.dom.css.Rect;
+
+import java.util.*;
+
+public class TextController {
+    private Map<Rectangle, String> texts;
+    private String nextMessage = null;
+    private BitmapFont displayFont;
+    private int screenWidth;
+    private int screenHeight;
```
BryanGraeser: It doesn't look like screenHeight is used

core/src/com/fallenflame/game/TextController.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public void update(PlayerModel player) {
+        if (texts == null) return;
+        for (Map.Entry<Rectangle, String> ele : texts.entrySet()) {
+            if (ele.getKey().contains(player.getX(), player.getY())) {
+                nextMessage = ele.getValue();
+                return;
+            }
+        }
```
BryanGraeser: So if I'm reading this right, do the player X and Y coordinates have to be exactly the value stored in order to transition the message? Having it required to be an exact match may be challenging for the player to actually trigger

core/src/com/fallenflame/game/TextController.java
```diff
@@ -0,0 +1,87 @@
+package com.fallenflame.game;
+
+import com.badlogic.gdx.Gdx;
+import com.badlogic.gdx.graphics.Color;
+import com.badlogic.gdx.graphics.g2d.BitmapFont;
+import com.badlogic.gdx.graphics.g2d.GlyphLayout;
+import com.badlogic.gdx.math.Rectangle;
+import com.badlogic.gdx.math.Vector2;
```
BryanGraeser: Doesn't look like this is used

core/src/com/fallenflame/game/TextController.java
```diff
@@ -0,0 +1,87 @@
...
+import com.badlogic.gdx.utils.JsonValue;
+import com.fallenflame.game.util.JsonAssetManager;
+import org.w3c.dom.css.Rect;
```
BryanGraeser: Or this

BryanGraeser (commented):
Few minor comments and one question: how precise does the player coordinate have to be to trigger the next message?

Also do you have a json with text as an example I can test this with?

core/src/com/fallenflame/game/TextController.java
foxhatleo: I will optimise the imports.

core/src/com/fallenflame/game/TextController.java
foxhatleo: It's not an exact value. The key is a rectangle so it is a box.

BryanGraeser (approved):
gtg

PR #122, opened 2020-05-05
App title

BryanGraeser:
Super short PR. Titles application "Fallen Flame"


foxhatleo (approved):
Looks good

foxhatleo:
@BryanGraeser I changed where title is set to DesktopLauncher instead of GDXRoot. This set titles more low-level so there is not that one brief second of "GDXRoot" as window title before changing to "Fallen Flame".

PR #124, opened 2020-05-05
Add fade-in fade-out transition

foxhatleo:
This PR adds a fade-in fade-out effect when switching among screens. The logic is implemented in GDXRoot.


BryanGraeser (approved):
Looks good

PR #125, opened 2020-05-06
Add key rebinding placeholder support for text

foxhatleo:
This PR adds support for key rebinding. Because key can be rebound, we don't want to include exact keys in texts because it might change. With this PR the text can be like:

> Press *{up}* to go up.

which will be replaced to

> Press W to go up.

Supported key placeholders are: `{up}`, `{right}`, `{left}`, `{down}`, `{reset}`, `{sneak}`, `{sprint}`, `{incLightRad}`, `{decLightRad}`


lsizemore8 (changes requested):
This does not work with the increaseLightRad and decreaseLightRad. I believe that is because in the LevelEditor, it's incLightrad while here it's incLightRad (and same with decLightrad vs decLightRad).

lsizemore8 (approved):
Looks good to go

PR #126, opened 2020-05-06
Pagified Level Select

stallop:
Level select now has pages, # of pages is uncapped atm as I don't think we know how many levels we'll end up having quite yet.


lsizemore8 (changes requested):
This would probably be better if we made the number pages based off of the length of the save.json (ie the number of levels total). This may be easier to edit w/ Bryan's branch, but get the length of the save.json array. and do the # of the pages on a array.length % 10 + 1, or something like that?

stallop:
> This would probably be better if we made the number pages based off of the length of the save.json (ie the number of levels total). This may be easier to edit w/ Bryan's branch, but get the length of the save.json array. and do the # of the pages on a array.length % 10 + 1, or something like that?

Good point, makes more sense than just hard coding it once we have a definite number of levels. Just made the change.

lsizemore8 (changes requested):
This looks a lot better. Just one last thing: don't show the "back" arrow unless we are on page 2 or more, and probably don't show the front arrow if this page shows the highest amount of **number unlocked** levels (so if the unlocked level is 25, don't show an arrow on the second page)

lsizemore8 (approved):
This looks good to me!

PR #127, opened 2020-05-06
Level progress and save

BryanGraeser:
Level progress and save functionality implemented!

The way this works is when a user first opens the jar, the contents of internal file `core/assets/jsons/save.json` will be copied to local file `savedata/save.json`. Local file `savedata/save.json` will then be updated as the player progresses and everytime they open the jar from then on, this local file will be used. 

The reason we need both an internal and local file is because the application cannot modify internal files, but the jar does not contain local files. Therefore, we need the internal file to initially create the local json file, and from then on we edit the local file.

I have pushed a version of the local file with all levels unlocked so we can access them all, but this **will not affect jar's** because I have changed the internal file to only have the first level unlocked. So any jar build will require you to unlock the levels.

**When adding a level, add it to the internal json `core/assets/jsons/save.json`, and make sure it is not unlocked. Then add it to the local save json `savedata/save.json` and make sure it is unlocked.**


BryanGraeser:
Leo's commits are here because I branched off of his branch originally (because it included changes to parts I was working on and I wanted to avoid alter conflicts)

ronnie-dumesh:
I reviewed the code and tested the functionality when building from IntelliJ. I also tested the jar and can confirm the game works fine using that. However, when I tried building my own jar the game crashes upon starting up, this might just be on my end. If somebody else can build a jar and just verify it works on a basic level, then the PR should be good to go

core/src/com/fallenflame/game/LevelSelectMode.java
```diff
@@ -1,8 +1,6 @@
 package com.fallenflame.game;
 
-import com.badlogic.gdx.Input;
-import com.badlogic.gdx.InputProcessor;
-import com.badlogic.gdx.Screen;
+import com.badlogic.gdx.*;
```
lsizemore8: Why *? Why are we importing so much?

BryanGraeser: That is a good question. Not sure how that happened, I blame Intellij lol.

Good catch

lsizemore8 (approved):
Looks snazzy and seems to work on my end (I created a new jar and everything)

ronnie-dumesh:
I tested rn and I was able to build a jar just fine in another branch but not in this one. I’m not sure if we wanna merge just yet

lsizemore8:
> I tested rn and I was able to build a jar just fine in another branch but not in this one. I’m not sure if we wanna merge just yet

That's really weird, I built the jar in this branch. When was the last time you created the build path (so did all of the steps to set up creating jars on the course website)

ronnie-dumesh:
Yesterday

PR #128, opened 2020-05-07
Tutorials complete

lsizemore8:
Creating this PR so people can see the tutorial levels, but all open PRs need to be merged in & need updates based on today's discussion before this one can be merged.


lsizemore8:
The Desktop/out is a result of an old t1 being merged in at one point into a jar. Can be ignored, as it'll be overwritten the next time one builds a jar

BryanGraeser:
When you say all open PR's need to be merged, can you tag which those are? Bc new ones are being added, so want to clarify which need to be in before I review/merge this

lsizemore8:
It's good, everything it was blocked on have since been merged in. It would be nice to test this with #126, but it's not blocked by it. @BryanGraeser it's all ready to be reviewed

BryanGraeser:
I don't see a way to play the tutorial levels. Also you seems to be having the same thing as Ravina where your local save.json file saves somewhere else

BryanGraeser (changes requested):
In the line stating "Beware, if that bar reaches zero...", "that bar" is very informal and not too helpful. It should be something like "Beware, if the bar on the top of your screen reaches zero...".

Also the hint in level 3 about sneaking comes after you've already evaded the enemy so it is not too useful. I don't think it's too big an issue bc you can evade the enemy without sneaking, but if the player is meant to follow that direction, it perhaps should be combined with the previous text direction.

BryanGraeser (commented):
For the save JSON stuff, this file shouldn't be here: `out/artifacts/fallen_flame_jar/savedata/save.json`, so if you could remove that from this PR that'd be great

Also why is there only one out level file `desktop/out/production/resources/jsons/levels/t1.json` does this need to be part of PR?

BryanGraeser:
I also unlocked all the levels on the local json that is in the other location just so you and ravina and anyone else whose OS makes it there will have them all unlocked

BryanGraeser (changes requested):
Also why does the tutorial still explain increase/decrease light radius? I thought we said we were getting rid of this

BryanGraeser (approved):

PR #129, opened 2020-05-08
Pause mode Null pointer fix

foxhatleo:
Can't really reproduce the null pointer so added a if statement to check for null values and a try-catch to make sure it does not fail.

### DIFF
--- a/core/src/com/fallenflame/game/PauseMode.java
+++ b/core/src/com/fallenflame/game/PauseMode.java
@@ -92,6 +92,7 @@ public class PauseMode implements Screen, InputProcessor {
             return false;
         } else {
             for (int i = 0, j = hoverRects.length; i < j; i++) {
+                if (hoverRects[i] == null) continue;
                 if (hoverRects[i].contains(screenX, screenY)) {
                     listener.exitScreen(this, i);
                 }
@@ -108,6 +109,7 @@ public class PauseMode implements Screen, InputProcessor {
     @Override
     public boolean mouseMoved(int screenX, int screenY) {
         for (int i = 0, j = hoverRects.length; i < j; i++) {
+            if (hoverRects[i] == null) continue;
             hoverStates[i] = (hoverRects[i].contains(screenX, screenY)) ? 1 : 0;
         }
         return true;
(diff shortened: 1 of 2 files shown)


core/src/com/fallenflame/game/PauseMode.java
```diff
@@ -107,8 +113,13 @@ public boolean touchDragged(int screenX, int screenY, int pointer) {
...
     public boolean mouseMoved(int screenX, int screenY) {
-        for (int i = 0, j = hoverRects.length; i < j; i++) {
-            hoverStates[i] = (hoverRects[i].contains(screenX, screenY)) ? 1 : 0;
+        try {
+            for (int i = 0, j = hoverRects.length; i < j; i++) {
+            if (hoverRects[i] == null) continue;
+                hoverStates[i] = (hoverRects[i].contains(screenX, screenY)) ? 1 : 0;
+            }
+        } catch (NullPointerException ignored) {
+            Arrays.fill(hoverStates, 0);
```
BryanGraeser: This is where it failed before, so this may be all we need. But what is the behavior here if `Arrays.fill(hoverStates, 0);` is executed? How does this impact the pause screen?

BryanGraeser (changes requested):
I think this is a little dangerous having it ignore all null pointers completely. It would make it very difficult to diagnose later issues. I commented the only when I think is necessary, and I also think for any try-catch we should at least have a println statement recording the error so if did need to figure it out later we still get some info.

core/src/com/fallenflame/game/PauseMode.java
foxhatleo: This basically reset all the buttons so that they are rendered as normal or "not hovered".

foxhatleo:
@BryanGraeser I see your point about the catch-all. I will adjust them.

core/src/com/fallenflame/game/PauseMode.java
lsizemore8: It's cause mouseMoved is an event-based state and the issue isn't hoverStates, but rather hoverRects, as that's an array of rectangles and is NOT filled before render. My guess is what's happening is if you move the mouse before render() completes one iteration, it crashes. Since mouseMoved is event-based, mouseMoved will take precedent over render, which is not what we want.

lsizemore8: I think this fix will actually fix the bug since I don't think we want to do an Arrays.fill() on hoverRects given what it's meant to do. The other option is to put an Arrays.fill() on hoverRects in the initialization of pauseMode

BryanGraeser (approved):
gtg now

PR #130, opened 2020-05-08
Add gradual light effect

foxhatleo:
This PR adds a transition for player light. A change in light radius will be rendered automatically with a rapid animation.

This does _not_ change the light radius in player model, so anything that relies on it (AI and whatnot) will use the real raw value. The transition is applied only in light controller when rendering the player light. However, the animation should be fast enough that it wouldn't cause an issue where the "light" AI sees and the light player sees feels out of sync.


BryanGraeser (approved):
gtg

PR #131, opened 2020-05-08
Golden master ui fixes

ronnie-dumesh:
the flare counter now works for any amount of flares and there is no more need for a global maximum in the flare counter as the last flare is drawn "first" and the rest to the left of it appropriotely. 

the sneak meter is also now adjusted in size appropriotely for any starting amount of sneak. This involved adding a method PlayerModel.getMaxSneakVal()


BryanGraeser (changes requested):
> the flare counter now works for any amount of flares and there is no more need for a global maximum in the flare counter as the last flare is drawn "first" and the rest to the left of it appropriotely.

What do you mean by it now works for any amount of flares? I tried enabling the player to pick up more than their max number of flares and it goes off the screen. Try running level 7 with my commit and pick up the first flare before firing (my commit allows the player to pick up a flare past their max). It is not drawn correctly in this case.

Sneak bar looks great though!

BryanGraeser (approved):
Okay, so currently discussing potentially allowing more than max flares. If we decide on that, it will come as a later PR. (Undid my change)

PR #134, opened 2020-05-10
Rebalance shooter

BryanGraeser:
Refactor fireball to instantaneous velocity and up speed to be more threatening. Also decrease sustained fire time.

Tested on level 3 and it seems fine. If a PR reviewer could give their feedback on bullet speed difficulty that'd be great.


ronnie-dumesh (approved):
I think the sustained fire time might be a bit too long, maybe instead of 50, try 35. However, if you believe 50 is good then I approve.

PR #135, opened 2020-05-10
Win lose screen

lsizemore8:
Created winning and losing screen based off of Ravina's designs. Couple notes:
1. The save data was accidentally pushed. I am tired, and will fix that Sunday morning. Creating this now so y'all can review the rest so I can make all changes at once
2. The title font is not officially in, as I'm currently trying to figure out if there is an actual legal/AI issue if we use it (it may only be available if we buy a commercial license). May not be in before the presentation as I may need some outside help to confirm this, but I figured it would be better to not let the font hold us up. 
3. The displayFont was made smaller in assets.json. This only impacts levelSelectMode, but it looked fine to me. If this is an issue, I can change the size w/ the code for that font for GameEngine, but I would rather not if I don't have to. I tried making two versions in the assets.json file for the different font sizes, but that did not work.


BryanGraeser (approved):
Looks good! One thing, "You won the game" sounds weird to me. Personally I think "You completed the game" or "You beat the game" sounds better, but this is obviously a personal opinion

PR #136, opened 2020-05-10
Size change

foxhatleo:
This PR adds resizability. The user can now resize the window or set the game fullscreen (tested on macOS Catalina).
It removes transition because there is a compatibility issue.


lsizemore8 (changes requested):
I tested this both on a 28inch monitor and my 13 inch Windows laptop. On both, performance was drastically hurt with this. On the large screen, the frame rate was about 10, and on the laptop, it maybe hit 15. I didn't really have anything open that could've caused this on my laptop. As cool as the resizing is, it's definitely not worth the frame rate drop. I'm not sure at this point if it's worth fixing, but it definitely can't stay as is.

PR #137, opened 2020-05-10
Enemy pause and exclamation mark

BryanGraeser:
Chasing enemies now pause upon activation and an exclamation mark appears above their head like so:

To change the duration of the pause time adjust `"totalPauseTime"` for enemy typeA in the global json.

Support could be added later for shooter enemy pause & exclamation mark if need be and would not be too challenging as I implemented this in a way to make it easier.


ronnie-dumesh (changes requested):
The enemy's animation stutters before it moves. Additionally, I think it makes sense that it have the same "pause" when you fire the flare, as logically it is "investigating". I don't see it making sense that the enemy instantly recognizes a flare but not the player. 

Finally, I think you should add the exclamation mark and pause support to the shooter as we discussed a sort of pause earlier. Some quick indication that it would start shooting would be very helpful. I think this change is very necessary. If we don't have time before tomorrow then lets place it on the backburner I guess.

BryanGraeser:
So I spent the past couple hours trying to get the shooter pause and flare pausing working and it isn't happening by tomorrow. They way the AIControllers are structured, flare pausing is going to be near impossible ever without a complete restructuring (bc I didn't know this would be a thing when we made it). Shooter pause should be working but is having some bugs I can't figure out in time for tomorrow so I'm going to make that a golden master goal.

Stutter is fixed

lsizemore8:
Given situation I think it's okay but @ronnie-dumesh you gotta approve it since you requested changes

BryanGraeser:
agreed

ronnie-dumesh (approved):
Approved

PR #138, opened 2020-05-10
Sound rebalance

stallop:
Minor changes to sound in line with the suggested "less is more" approach. Enemy constant noise and bgm values turned down a bit, and I implemented a threshold constant that gets subtracted from all enemy constant noises. This essentially just silences very quiet distant enemies by putting their value below zero, so not all enemies are always making noise.


lsizemore8 (changes requested):
Could you make the flare sound a tad quieter as well? That one is quite loud in comparison. Past that it looks good

stallop:
> Could you make the flare sound a tad quieter as well? That one is quite loud in comparison. Past that it looks good

Just did it.

lsizemore8 (approved):
Looks good

PR #139, opened 2020-05-10
Added a new level

BryanGraeser:
Structured like this: 

Starting flares: 3. Meant to force the player to pick up more flares


BryanGraeser:
Please play it and lmk what you think

ronnie-dumesh:
really like the level overall! The new shooters are a lot bigger so you'd have to put some more distance between the flare and the shooter in the lower-left hand side of the map. I also thought that the stationary enemy in the top-left felt a little cheap. I was trying to avoid the moving enemy and thought I was all free after I flared that one and then the stationary one got me cuz I couldn't tell whether it was there due to the technical limitations of the fog. Consider getting rid of that one but your call 

Also, I beat it without using sneak/sprint at all. That's fine I suppose but not sure if you'd want to force it a little more

Finally, there's a print statement being printed that says "start not safe", not sure what that's about and I couldn't find it in the committed code

ronnie-dumesh:
Also as reference for anyone else testing it's level number 7

ronnie-dumesh:
Also, I beat it without using sneak/sprint at all. That's fine I suppose but not sure if you'd want to force it a little more

BryanGraeser:
Yeah, I wanted this level to be harder hence that stationary enemy, but having it so close to the end may be more frustrating than challenging.

I know what's printing that--it's the pathing and I think it relates to some of the enemies running into walls issues we're having. I can look into it

BryanGraeser:
Okay, should be gtg

PR #140, opened 2020-05-10
changed sprint light radius

stallop:
It's literally just changing a single value in global.json. Idk if I should have made an entire PR just for this, but figured it wouldn't hurt and is pretty easy to check off on.


ronnie-dumesh (approved):
Approved

PR #141, opened 2020-05-10
added new level from Max

stallop:


lsizemore8 (approved):
Resolve the conflict but I like it!

core/assets/jsons/save.json
```diff
@@ -1,50 +1,57 @@
...
+  "levels": [
+    {
+      "name":       "easyLevel",
+      "unlocked":   true,
+      "completed":  true,
+      "path":       "levels/easyLevel.json"
+    },
+    {
+      "name":       "mediumLevel",
+      "unlocked":   true,
```
BryanGraeser: This file should NOT be set to all unlocked. This determines what is initially unlocked in the JAR

BryanGraeser (changes requested):
`core/assets/jsons/save.json` should not have all files unlocked.

core/assets/jsons/save.json
BryanGraeser: You should add your level, but leave everything locked except the first tutorial level

BryanGraeser (approved):
fixed save issues. gtg

PR #142, opened 2020-05-11
Golden master texture fixes

ronnie-dumesh:
Added the new textures for the shooter and the bullet it fires and changed physics bodies respectively. I also replaced references to the "sneakMeter" with "powerMeter" and now sneaking AND sprinting cause it to go down. Finally, I added a death animation that triggers if the player is touched by the enemy. The player can still move for a moment, but I feel that this is fine for realism and don't see any real issues with this.


core/src/com/fallenflame/game/LevelController.java
```diff
@@ -1050,7 +1053,7 @@ public void beginContact(Contact contact) {
             // Check for loss condition 1 (player runs into enemy)
             if((bd1 == player && bd2 instanceof EnemyModel)
                     || (bd1 instanceof  EnemyModel && bd2 == player)){
-                setLevelState(LevelState.LOSS);
+                player.die();
```
lsizemore8, line 1075: It'll probably be worthwhile to, instead of doing this, send up a state (maybe LevelState.LOSING) that, if true, doesn't let the player move. It does feel really strange that I can sprint my character away as it falls to the floor. Either that or the animation needs to be faster, but I think stopping all movement would feel better.

lsizemore8 (changes requested):
I would say that's too long. Either stop all movement (update the enum to have a 4th state, LevelState.LOSING and update GameEngine to, if this state is true, do nothing when the player tries to do any action. Also, make sure you continue to merge master into here, there's been a lot of changes, esp with regards to win/lose state

core/src/com/fallenflame/game/LevelController.java
ronnie-dumesh: So, I made a one line fix that essentially stops the player from moving. When the player is hit by any enemy they now cannot move, however, the player can still be pushed by the walking enemy a very short distance (maybe at most 1m). If any other enemy hits the player they stop on a dime. I think it's ok rn.

core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -406,7 +406,11 @@ public void update(float delta) {
             // If player just stopped sneaking
             level.makeWalk();
         }
-        level.getPlayer().move(moveAngle);
```
lsizemore8: Can you also stop it from creating flares & sprinting/sneaking when they die? Ideally it would stop enemy movement too (so like the only thing that would happen is the death animation once the player is dying), but flares/sneaking/sprinting shouldn't be too hard

lsizemore8 (changes requested):
Definitely a lot better. It's a tad strange that the enemy pushes the player around so ideally enemy movement is stopped too. What would be more important, however, is stopping flares/sneak/sprint as well. That should be easy, but while I'm dying I can trigger the ghost rn, which isn't great.

core/src/com/fallenflame/game/GameEngine.java
BryanGraeser: Is there a way to freeze the world? This is just a hunch, but perhaps not calling world.step()? Not sure if that would break the death animation but something to try

BryanGraeser: FYI that happens here: 
`world.step(timeStep, WORLD_VELOC, WORLD_POSIT);`
in function fixedStep in LevelController

lsizemore8 (approved):
In the future we definitely need to change the minimum sneak/sprint, but this looks good enough that I'm approving

PR #143, opened 2020-05-11
Laura level

lsizemore8:
image


BryanGraeser:
JSON stuff looks good. Haven't playtested yet, but just wanted to comment that

BryanGraeser:
Played it and it seems fine. Tbh is a little easy, but we could just have it as an earlier level (also it may be bc I've played our game so much lol)

lsizemore8:
Yeah that was intentional. I wanted an easier level to be added so we can have a greater range. Will deal with the merge conflicts now/merge it if you can approve it

BryanGraeser (approved):
Technically looks gtg (I resolved merge conflicts). I think it is maybe a little easy and also you call it "force sneak," but if this was the intent I don't think it forces sneak (you can just fire a flare right at the start). If you want to force sneak, maybe start the player with no flares and have them pick them up in the middle of the level, this would also increase difficulty and add another level that uses pickups.

BryanGraeser:
Your choice if you want to make any level design changes before merging

lsizemore8:
Oooh good point. Yeah I'll update that after my next class.

BryanGraeser:
Talked with Laura, merging now and later PR for level design changes.

PR #144, opened 2020-05-11
Ronnie level

ronnie-dumesh:
#Added the ability to specify a different number of start and end flames. 

I couldn't figure out how to have my level be playable by adding it to the end of save.json. Right now it is currently the first level. If someone can help me fix this that'd be appreciated. 

The idea behind this level is that you need to pick up flares and try to finesse your way through either the upper or the lower part of the level. 

[Screenshot from Gyazo](https://gyazo.com/2755fec83aeb4d6b22c738ec05d02458)


core/assets/jsons/global.json
```diff
@@ -238,7 +238,7 @@
         "y": 0.30
       }
     },
-    "maxsneak": 1000
+    "maxsneak": 500
```
BryanGraeser, line 244: Why? Is this to change the sizing of the bar?

BryanGraeser (changes requested):
I have one question, mentioned below. And I have one level design suggestion: define a "startSneakVal" that is lower than the default. It is quite high for this level, and in general I don't think we should use the default value (it is really only there for backwards compatibility).

BryanGraeser:
I also dealt with the save stuff and resolved merge conflicts

BryanGraeser (approved):
nevermind dealt with it myself bc we need to merge

PR #146, opened 2020-05-11
add throwing animation

ronnie-dumesh:
Add throwing animation to fire buddy. Animation does not reset if you throw multiple during its duration.


BryanGraeser (approved):
Play-tested and it looks good to me

PR #147, opened 2020-05-11
balancing changes

BryanGraeser:
- Removed flares from first few tutorial levels
- Removed flares from last non-test level bc otherwise it's crazy easy
- decreased pause time
- set sneak values on a couple levels that didn't have it


lsizemore8 (approved):
gtg

PR #149, opened 2020-05-12
fixed my level to have no flares

lsizemore8:
Made my level more challenged based off of feedback from Bryan.


BryanGraeser (approved):
gtg

PR #150, opened 2020-05-13
add delay to end of player death animation

ronnie-dumesh:
Adds a number of frames to the end of the player death animation before the player is considered "dead" by the game and the "game over" screen pops up

### DIFF
--- a/core/src/com/fallenflame/game/PlayerModel.java
+++ b/core/src/com/fallenflame/game/PlayerModel.java
@@ -73,6 +73,9 @@ public class PlayerModel extends CharacterModel {
     private FilmStrip deathFilmstripRight;
     private FilmStrip deathFilmstripLeft;
 
+    /** Frames until player finishes dying and is considered dead */
+    private int deathDelay;
+
     /**
      * Initializes the character via the given JSON value
      *
@@ -106,6 +109,8 @@ public class PlayerModel extends CharacterModel {
         life = LifeState.ALIVE;
 
         throwing = false;
+
+        deathDelay = globalJson.get("deathdelay").asInt();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


core/src/com/fallenflame/game/PlayerModel.java
```diff
@@ -231,9 +236,10 @@ protected void setFireBuddyOrigin(float x, float y){
     /**
      * Get the position of the firebuddy on the screen in meters
      */
-    public Vector2 getFireBuddyPosition() { return new
-            Vector2(getPosition().x + (getFireBuddyOriginX() + fireBuddyFilmstrip.getRegionWidth()) / drawScale.x / 2.0f,
-            getPosition().y - (getFireBuddyOriginY() / drawScale.y) + fireBuddyFilmstrip.getRegionHeight() / drawScale.y / 2.0f);}
+    public Vector2 getFireBuddyPosition() { return new Vector2(
```
lsizemore8, line 239: Why is this changed here?

core/src/com/fallenflame/game/PlayerModel.java
```diff
@@ -435,10 +439,14 @@ public void update(float dt) {
...
                 walkCool--;
-            } else if (frame == filmstrip.getSize() - 1){
+            } else if (deathDelay <= 0){
                 life = LifeState.DEAD;
+            } else if (frame == filmstrip.getSize() - 1) {
+                deathDelay--;
             }
+
         } else if (isAlive()) {
+            animateFireBuddy(angle100);
```
lsizemore8, line 449: Also why is this changed?

BryanGraeser:
Other than Laura's comments, it looks good

core/src/com/fallenflame/game/PlayerModel.java
ronnie-dumesh: I tried to factor things out into variables but then realized it would look even worse than it does now. There's no differenec in the code except I just changed the line of the initializer

core/src/com/fallenflame/game/PlayerModel.java
ronnie-dumesh: Fire Buddy would "animate" while player was dying. It would essentially just keep shaking side-to-side. Here, it does not move at all while player is dying

lsizemore8 (approved):
Changes make sense. Looks good.

PR #151, opened 2020-05-13
change ghost spawn area to be closest boundary

ronnie-dumesh:
The ghost now spawns the at boundary wall the player is closest to. I also reduced the ghost's speed to give the player a chance to escape the ghost while forcing them to keep moving.


core/assets/jsons/global.json
```diff
@@ -126,7 +126,7 @@
       "density": 1.0,
       "friction": 0.0,
       "restitution": 0.0,
-      "walkspeed": 	  4.0,
+      "walkspeed": 	  2.4,
```
lsizemore8: The ghost is too slow now. If I managed to get ahead of the ghost, I can literally walk at any speed I want (even sneak a little) and not feel threatened. In short, it almost nerfs the ghost on maze levels. I don't think it should be as fast as before though. I think maybe a speed of 3.0 may suffice.

lsizemore8:
So issue: This fix now creates a new problem: getting the ghost as you are about to win. It makes the game really frustrating, as the ghost now appears in the exact location you want to go, so now you have to essentially restart just as you were going to win as opposed to before, where the earlier you were in the level the more the ghost hurt you, therefore reducing frustration. This is especially apparent in a level like level 8, where the optimal strategy (sneak through it) is literally impossible now. In essence, any sort of level where the optimal strategy at the end is to sneak through it is just very frustrating and not fun. I think what would be better is the ghost to come from the opposite boundary (or opposite point that you can see) and the ghost moves faster.

lsizemore8 (changes requested):
See comments. This fixes one issue, but creates a new one instead

BryanGraeser:
Discussed in meeting: New desired functionality is ghost will appear opposite the exit on the line between the player and the exit. Distance from player will be a constant ratio of player distance to exit

core/src/com/fallenflame/game/LevelController.java
```diff
@@ -802,6 +815,31 @@ public void addGhost() {
...
+
+        float ghostX = player.getX() + (dx * ghostSpawnScale.x);
+        ghostX = angle > pi / 2 && angle < 3 * pi / 2 ? ghostX - ghostSpawnOffset.x : ghostX + ghostSpawnOffset.x;
+
+        float ghostY = player.getY() + (dy * ghostSpawnScale.y);
+        ghostY = angle < pi ? ghostY + ghostSpawnOffset.y : ghostY - ghostSpawnOffset.y;
+
+        return new float[]{ghostX, ghostY};
+    }
+
```
BryanGraeser: This method header is no longer accurate, please update to reflect the function

BryanGraeser (changes requested):
I resolved merge conflicts so I could review the PR. I still think the ghost is too slow and may spawn too far away. For example on level 6, if I spawn the ghost right at the start, it takes a very long time to show up. This might be confusing for the player, because they see the bar run out, yet nothing happens for a while, so they may think something else finally triggers the ghost. 

I think in general just adjust the ghost constants making it spawn closer and move faster. Using up all the sprint/sneak is meant to trigger a frantic, endgame dash to the exit, but it doesn't do that right now.

ronnie-dumesh:
I made the changes requested. If you don't like how far the ghost is you can change the values of scale. I think it's good as is and essentially just comes to fiddiling/UI.

BryanGraeser (approved):
I still think the ghost should go a little faster, but that's just my opinion. Otherwise gtg.

lsizemore8:
I think it's too slow. In most cases, I can now just sneak to the exit. So even if the ghost appears, I still have plenty of time to make it, while sneaking the entire way on some levels (provided I'm not right at the beginning).

lsizemore8 (approved):

lsizemore8:
Approved but make the ghost faster

PR #153, opened 2020-05-14
changed local to external

lsizemore8:
Due to the packr not working on Mac, I updated GameEngine to use external instead of local.


lsizemore8:
@ravinapatel, please test this by seeing if your local save.json is updated in the correct folder (so not the savedata in core/assets)
@ronnie-dumesh, please test this by creating a jar and opening it by clicking, you are the only one who hit the bug w/ the level unlocking, so you need to test this.

BryanGraeser:
Where is this file written? I can't find it

BryanGraeser:
I literally cannot find it. I deleted every `save.json` I could find and it's still running, so idk where it is.

BryanGraeser:
Ok nevermind, it goes to your $HOME directory. @lsizemore8 can you remove all the old outdates save.json files and savedata folders now? And I think this should be tested by mac users before merging.

BryanGraeser (approved):
gtg (assuming someone on Mac tests)

Issue #73, opened 2020-04-13
Fog does not disappear if enemy is activated


lsizemore8:
This is a bigger issue with regards to the shooting enemy. But the fog particles appear if the shooting enemy exists, which doesn't make much sense.

PR #154, opened 2020-05-18
Showcase fog fixes

lsizemore8:
Fixed/improved black circle issue around player and flare lights. Couple notes: 
1. This branch assumes that we will be removing the light radius w/ scroll wheel soon, so it does NOT work well if you increase the light radius too much w/ the scroll wheel. 
2. The flares are a tad strange, because w/ the player, the dark circle is counted as a "trigger" area and can activate the enemies, while with the flares, that doesn't seem to be the case. Those testing it, try the flare with level 5, you'll notice that it now sometimes illuminates the enemy without activating it. I left it in so people could see, but it's a quick change back to the black circle for flares (or I can change the fog to cover the black circle for flares easily) if that is a bad feature. 
3. It's a hacky solution. I fully admit it. The other option (fixing the diffuse) seemed like more time than it was worth.


lsizemore8:
Converted to draft because I realized that at some point in the past few weeks, I accidentally deleted the code that removes fog when an enemy is activated... trying to fix now/see if it's worth adding back

lsizemore8:
Fixing the enemy issue is gonna take many hours of work based on how we did this, so I give up, and will fix it in a later PR if I have time

BryanGraeser:
Is there anyway to get the fog closer to the player? There still is a bit of black circle around them, and it looks a little funky when you walk. Though it is definitely not the biggest issue and other fog-related stuff should be prioritized.
image

core/src/com/fallenflame/game/FogController.java
```diff
@@ -72,9 +72,10 @@ public void updateFogAndDraw(GameCanvas canvas, Vector2 scale, float delta) {
                 }
                 //To prevent drawing on tiles with the player or a wall as well as if its within the light radius
                 if (levelModel.hasWall(x, y)) continue;
+                //0.5 accounts for the aligning of the light to show the player's face over the feat.
```
BryanGraeser: What does "show the player's face over the feat" mean?

core/src/com/fallenflame/game/LightController.java
```diff
@@ -128,15 +131,15 @@ public void initialize(PlayerModel player, ExitModel exit,
...
         rayhandler.setBlurNum(3);
-
         updateCamera();
 
         // Save player and config.
         this.player = player;
         this.lightingConfig = levelLighting;
-
+        playerLightOffset = (player.getLightRadius()/2) * 0.4f;
+        flareLightOffset = 0.5f;
```
BryanGraeser: Why are constants being defined in initialize()? You should just make both of these vars constants defined above.

BryanGraeser (commented):
There still seems to be a black circle around the flare too
image

BryanGraeser:
The two issues I mentioned with there still being a black circle probably just mean you need to tweak the constants some more (and I think having a bit of fog lit up is preferable to having a bit of black edge).

BryanGraeser:
> 2. The flares are a tad strange, because w/ the player, the dark circle is counted as a "trigger" area and can activate the enemies, while with the flares, that doesn't seem to be the case. Those testing it, try the flare with level 5, you'll notice that it now sometimes illuminates the enemy without activating it. I left it in so people could see, but it's a quick change back to the black circle for flares (or I can change the fog to cover the black circle for flares easily) if that is a bad feature.

I don't think this is an issue you introduced. I just tested on master & the enemies can be triggered by the "blackness" outside the light, so the radius of the light should correspond to activation, it just may be slightly off --> this may indicate this issue is more on the fog not being close enough than the light not being large enough.

A possible solution to this is to have a similar constant added to where enemy flare detection is handled.

core/src/com/fallenflame/game/FogController.java
lsizemore8: It accounts for the fix Ronnie did way back w/ the light radius not showing the players face. If you look at an old branch, you'll notice the black circle is offset. I didn't make it a constant cause it's used once.

core/src/com/fallenflame/game/LightController.java
lsizemore8: The flare one can be a constant, but the player one cannot because it needs to change based on the current value of the light radius.

lsizemore8:
With regards to black circles, it's an easy fix so I can do that, as it's just changing the values. I can make some of the values (the multiplier) a constant, but not the offset itself, as it's not a constant. I'm using this branch to fix the bigger fog issue, so those fixes will come with that fix, if I can ever get it done.

core/src/com/fallenflame/game/FogController.java
BryanGraeser: Oh do you mean "feet"?

core/src/com/fallenflame/game/LightController.java
BryanGraeser: I meant more the flareLightOffset and the multiplier for playerLightOffset, but it's not the biggest deal if you just leave the playerLightOffset as is

core/src/com/fallenflame/game/FogController.java
lsizemore8: ...yes

lsizemore8:
Finally, finally ready for review. This fixes issue #73 in addition to the black circle issue.

BryanGraeser (changes requested):
One major issue this has is the player can now fully illuminate an enemy without activating them:
image

So there is still some significant issue btw light radius and activation radius.

Further, and this is less significant, there is an issue with almost 2 rings of light. First the inner lit region without fog, and then a ring around that with fog but still different lighting. It also flickers as you walk. I understand this may be the cost of getting rid of the black region, but is it possible to at least shrink this band?

BryanGraeser:
The second "issue" is not as big of a deal, but the first one is in my opinion, game-breaking

BryanGraeser (approved):
I think these changes are looking good to go! The enemies leave a black line of now fog sometimes when they walk when activated, but I think this is not too much of an issue.

Issue #123, opened 2020-05-05
Enemy still gets stuck in corner level 2


lsizemore8:
If you sprint for the first part of the level, the enemy will get stuck in the corner and you can walk past

![image](https://user-images.githubusercontent.com/32024811/81075955-0740c280-8eb9-11ea-9260-84ec14fb3c62.png)

BryanGraeser:
Probably an issue with the grid for pathfinding

PR #156, opened 2020-05-19
Fix pathfinding bug

BryanGraeser:
As discussed with @lsizemore8, this bug is a result of the grid sizing for pathfinding, however fogController is now also dependent on this sizing. I have thus refactored levelmodel to maintain fog and pathfinding grids separately (to allow different sizing constants for both).

This solves issue #123 without breaking our fog implementation.

### DIFF
--- a/core/src/com/fallenflame/game/enemies/AIController.java
+++ b/core/src/com/fallenflame/game/enemies/AIController.java
@@ -93,60 +93,60 @@ public abstract class AIController {
      * @return a movement direction that moves towards a goal tile or NO_ACTION.
      */
     protected int getMoveAlongPathToGoalTile() {
-        int startX = level.screenToTile(enemy.getX());
-        int startY = level.screenToTile(enemy.getY());
+        int startX = level.screenToTile(enemy.getX(), true);
+        int startY = level.screenToTile(enemy.getY(), true);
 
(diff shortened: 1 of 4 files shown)


core/src/com/fallenflame/game/enemies/AIController.java
```diff
@@ -93,60 +93,60 @@ public int getAction(){
...
      */
     protected int getMoveAlongPathToGoalTile() {
-        int startX = level.screenToTile(enemy.getX());
-        int startY = level.screenToTile(enemy.getY());
+        int startX = level.screenToTile(enemy.getX(), true);
+        int startY = level.screenToTile(enemy.getY(), true);
 
-        if(!level.isSafe(startX, startY))
-            System.out.println("start not safe");
+//        if(!level.isSafe(startX, startY))
```
lsizemore8, line 99: Why is this commented out as opposed to just removed?

core/src/com/fallenflame/game/enemies/AIController.java
```diff
@@ -93,60 +93,60 @@ public int getAction(){
...
+//            System.out.println("start not safe");
```
lsizemore8, line 100: Same here

BryanGraeser:
Better fix: #158

PR #157, opened 2020-05-19
Remove light scrolling from player; make debug option

BryanGraeser:
- Removed the light scroll wheel as an option usable by the player
- Updated edit control screen to reflect that
- Kept light scrolling as a debug option --> ie when `ALLOW_DEBUG = true`, scroll wheel and comma/period still work

### DIFF
--- a/core/src/com/fallenflame/game/GameEngine.java
+++ b/core/src/com/fallenflame/game/GameEngine.java
@@ -51,7 +51,7 @@ public class GameEngine implements Screen, InputProcessor {
 
     /**Boolean to determine if debug keys do anything.
      * Note: MUST BE FALSE WHEN MAKING A JAR! */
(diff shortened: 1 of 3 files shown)


core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -51,7 +51,7 @@
 
     /**Boolean to determine if debug keys do anything.
      * Note: MUST BE FALSE WHEN MAKING A JAR! */
-    private static final boolean ALLOW_DEBUG = false;
+    private static final boolean ALLOW_DEBUG = true;
```
lsizemore8: This should never be set to true on master, set this to false

lsizemore8 (approved):
Works well. Tried my best to break it without fail

PR #158, opened 2020-05-19
Fix pathfinding bug V2

### LINKED ISSUE #123: Enemy still gets stuck in corner level 2

BryanGraeser:
Cleaner fix than #156 

As discussed with @lsizemore8, this bug is a result of the grid sizing for pathfinding, however fogController is now also dependent on this sizing. **For this fix I have just created two levelModels: one for fog, one for pathfinding**. This could arguably have a bit higher overhead than fix #156, but only marginally so and this solution is much cleaner code and architecture wise.

Closes #123.

### DIFF
--- a/core/src/com/fallenflame/game/FogController.java
+++ b/core/src/com/fallenflame/game/FogController.java
@@ -9,8 +9,6 @@ import com.badlogic.gdx.utils.Array;
 import java.util.Iterator;
 import java.util.List;
 
-import static com.fallenflame.game.LevelModel.TILE_SIZE;
-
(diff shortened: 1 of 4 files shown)


lsizemore8 (approved):
Is it still relatively hacky? Yeah. Is it better than trying to redesign fog? Most definitely. Seems to work well and cleaner than the prior fix. Only difficult thing is going to be remembering to update both, but that hopefully won't come up.

PR #159, opened 2020-05-19
Enemy shooter pause

BryanGraeser:
Adds pause effect to shooter enemy when player light radius hits it (just like moving enemy). 

Notes:
- Updated light controller constant for speed of enemy light radius expansion upon activation to be faster (issue was by the time it expanded, the enemy was basically done pausing)
- Kept pause time short to maintain difficulty
- Does not pause on flares (may add this feature later if time)
- Does not require any changes to level json's

Please playtest and lmk what you think!


stallop (approved):
Functions very well, and the faster light expansion is a nice touch. Although I know you want to keep the pause time short because this is already a pretty easy enemy to avoid, I feel like sometimes the pause is hardly even noticeable. If it's necessary to keep the enemy from being too easy though, I think it should be fine.

BryanGraeser:
> Functions very well, and the faster light expansion is a nice touch. Although I know you want to keep the pause time short because this is already a pretty easy enemy to avoid, I feel like sometimes the pause is hardly even noticeable. If it's necessary to keep the enemy from being too easy though, I think it should be fine.

Playtested some more, and I agree it is only there for a second, but I think otherwise it'll be too easy. It'd be an easy change though if playtesting indicates otherwise.

PR #160, opened 2020-05-20
Shooter prioritizes player over flares

BryanGraeser:
Shooter will now prioritize shooting the player over flares, so if it sees both a player and a flare, it will shoot at the player. **If it just sees a flare first it will shoot at it without pausing.** This means:
- If it sees the player first, it will pause (even if it sees a flare after)
- If it sees a flare first it **will not pause** and start shooting at the flare (so if it sees the player next it will immediately, without pausing, switch to shooting them)  
- If a shooter is doing sustained fire and sees the player, it will not pause and immediately shoot at them

Basically in short, **the shooter only pauses if it sees the player first from it's idle state.** The idea is once the shooter is "activated" it will not pause until it is "deactivated," and flares will **never** pause the shooter. This is meant to make the shooter more of a "punishment"/downside to using flares. 

What do you think of this behavior?

*Sidenote: ignore the branch name*


BryanGraeser:
Dependent on #159 (branched off of that to prevent later merge conflicts)

BryanGraeser:
Should be ready for review/merge now.

stallop (approved):
Functions well and I think it's a good way of balancing things. Like you said, it creates additional risk to shooting flares, and additionally it helps buff the shooter into a more formidable enemy.

PR #161, opened 2020-05-20
Make sprint deplete resource bar faster

BryanGraeser:
Sprint now depletes resource bar 2x as fast as sneak.

Note: In general, levels going forward should (and will) have a high resource value.

### DIFF
--- a/core/assets/jsons/global.json
+++ b/core/assets/jsons/global.json
@@ -15,6 +15,8 @@
     "sneaklightrad":    1,
     "minlightradius":   2,
     "defaultStartSneakVal":   300,
+    "sprintDecRate": 1.4,
+    "sneakDecRate": 0.7,
     "texture":     	{
       "left": "player-left",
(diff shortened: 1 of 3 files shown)


ronnie-dumesh:
Could you make this data-driven so we can just change some sort of constant in global.json? We could potentially balance this some other way. We might also want to make the power value a float now since we could decrease by like 1.5

BryanGraeser:
I wanted to but currently it is an int, not a float. I can change it to a float though

BryanGraeser:
Changed it to a float and made it constant-based. Also decreased both rates slightly

ronnie-dumesh:
Could you make it as part of global.json so it can be modified there?

BryanGraeser:
> Could you make it as part of global.json so it can be modified there?

Done

ronnie-dumesh (approved):
Approved

PR #163, opened 2020-05-21
add winning animation and delay to victory screen

ronnie-dumesh:
Also replaces player-walk-left and player-walk-right textures


lsizemore8 (changes requested):
Couple issues: 
1. If I sprint to the exit, and then let go of sprint during the animation, the light goes back to the initial value pre-sprint
2. If I continue to hold down one of the WASD (potentially others too), the animation speed slows down drastically.

ronnie-dumesh:
1. Resolved
2. I made it so animation speed is always constant now regardless of WASD inputs. It's a bit slower than originally but I think it looks better that way

lsizemore8 (approved):
To save time I fixed the sound issue real quick. Looks gtg, but make sure you merge master in first (after my fix is approved/merged)

PR #164, opened 2020-05-21
World select

stallop:
World select (hopefully) working.

Note: To test all the functionalities, I have the save data currently put one level in world 2, which is unlocked, and two levels in world 3, which are both locked. I've been trying different combinations to make sure unlocking worlds, unlocking levels, and pagination all work together properly, but might have missed some. Feel free to mess with the save json by moving levels around to different worlds and locking/unlocking them to see if there's any cases the code doesn't properly handle.

PR #165, opened 2020-05-21
updated backgrounds and win/lose with new fonts

lsizemore8:
Updated menu, loading, credits, and the win/lose UI to have the new designs (ie the new text). Also fixed the issue where the border around the win/lose screen was always set to the border for the levels you did not win the entire game. Did NOT update level select, as that will be done in #164. Also removed a few of my old art, because I didn't want to see it anymore in the folder.


BryanGraeser:
I think we should close this because of #166

PR #166, opened 2020-05-21
UI updates

BryanGraeser:
UI updates (excluding level select which Max is working on)


lsizemore8:
Writing this here so I can fix it in the morning. Menu screen needs to be updated: Start and Select should be combined to just be "play"

ravinapatel (approved):

BryanGraeser (commented):
I think we need to add 'pause' to the controls screen before merging

lsizemore8 (changes requested):
If you change Pause to something else, you then cannot change it back to 'esc'

BryanGraeser:
Contemplating changing pause to 'p' and never allowing 'esc' pausing

lsizemore8:
If you press "reset to default", pause goes back to being escape instead of p

lsizemore8 (approved):
Looks gtg

PR #167, opened 2020-05-21
Wall fix

ronnie-dumesh:
This PR does one three things

1. modifies the double-wall-side-view and double-wall-top-view to be 64 x 64 instead of 65 x 65 and removes any padding these walls had
2. removes unused textures
3. modifies level/t1 to fit in line with the requirement that wall textures partially instead of fully overlap


foxhatleo (approved):
gtg

PR #168, opened 2020-05-21
New world assets

foxhatleo:
This PR adds new world assets and is compatible with current level editor JSON implementation. Specifically it adds:

- Two new background texture ("forest-tile" and "volcano-tile")
- A new set of wall textures ("wall-side-volcano" and "wall-top-volcano")
- Trees


packr-stuff/packr-mac.json
```diff
@@ -5,6 +5,7 @@
     "classpath": [
         "fallen-flame.jar"
     ],
+    "icon": "icon.icns",
```
BryanGraeser, line 8: Why this change? Is this adding the fallen flame icon?

BryanGraeser (changes requested):
Looks really good! I have a few major questions though:

1. The tree physics body should be a big larger, and I think it perhaps should also be square. I understand you don't want to to feel too artificial, but right now it's size is a bit too conservative and this combined with the circular nature makes it confusing to walk along. For example you can get caught in positions like this, and I can foresee some frustration on the part of the player
image
I get the need for drawing in a specific order bc the branches go higher than where the physics body should be, and I think this is handled well. However, is there any real need for it to be circular? If we do make it square, I think the level-editor should show the square physics body of the tree so it is easier to set and it should snap to tiles like the walls. This will keep our level-design process and components consistent. In short: I think it should behave more like walls.

2. Flares currently go through trees. Discussing this in slack rn, but idk if we want this.

3. I think the border for forest levels should be trees, not walls as it currently is in the level-editor. This is obviously more of a level-editor change, but I thought I'd mention it here.

BryanGraeser:
The more I think about, the more I think trees should just act like wall objects (if not be wall objects) where the texture extends higher than the physics body

packr-stuff/packr-mac.json
foxhatleo: Yeah this is a packr update. I told Laura i put it in this PR to save time cos it's a tiny update.

BryanGraeser (approved):
gtg

PR #169, opened 2020-05-21
Texture hotfix

lsizemore8:
Added back the slime texture, as slime should not have been deleted in the first place, and removed the logic that was calling the inactive_flare_counter


core/assets/jsons/global.json
```diff
@@ -254,8 +254,7 @@
   },
   "flarecount": {
     "texture": {
-      "active": "activeflarecounter",
-      "inactive": "inactiveflarecounter"
```
BryanGraeser, line 258: Can you remove `inactiveflarecounter` from assets too so it isn't being loaded?

BryanGraeser (approved):
One small change, then gtg

core/assets/jsons/global.json
lsizemore8: It's already removed

PR #170, opened 2020-05-21
Tutorials update

lsizemore8:
Added in the new tutorial levels, and made changing levels by number (a fix added for pre-beta, before level select was a thing) only available when ALLOW_DEBUG is true.


BryanGraeser (changes requested):
Definitely some big improvements here, but there's still a decent number of changes to be made.

Tutorial Level 4 (flare) still has a lot of issues:
- The **big** one that TA's have been complaining about: This level is supposed to teach you about flares but does not require you to actually throw a flare. Like I mentioned after we talked to the TA, the level should be: a long hallway that is narrow enough an enemy blocks it, there is an enemy between the player and the exit, but there is space past the exit in the hallway. Ergo, the solution is to throw a flare down the hallway so the enemy goes past the exit.
- I think the first text "Aim with the mouse..." should show up right away instead of you having to walk forward to prompt it. 
- I also think you can get rid of the last text in that level "Walk to the exit" bc there's already a lot of text here with little time to read.

Level 5 (LShift) issues:
- Again this level does not actually require the use of the ability it is teaching. I successfully used sneak to pass the enemy instead of L-Shift lure like I was supposed to. The gap around the enemy needs to be tighter.
- the instructions need to mention that enemies go to your last known position. Also I suggest replace "let go" with "release"

Level 6:
- Small change, but I think the corridor at the end should be longer so when it says "run" you actually have a bit of frantic running. This is more of a "feel" change.

Finally, **there is still no shooter tutorial.** As the TA mentioned this **needs** to be added.

BryanGraeser:
Text on level 4 is way too long (also last sentence is missing a period).
Shorter version:
> An enemy is blocking your path! It will only chase you if your light touches it. Hold L-Shift to lure it towards you, and release to sneak by. The enemy will go to your last known location. 

For the shooter level, the initial text disapears way too fast. Extend the level space if you need, but it needs to stick around longer.

BryanGraeser (approved):
gtg

PR #171, opened 2020-05-21
Wall flare

stallop:
Works but had to change things slightly in other parts. For some reason timeToBurnout() was using 1 to indicate is wasn't on a wall yet, and anything calling it was just checking if it was <= 0 to make sure it was burnt out. This was causing a pop at the 1ms mark on fading, so I changed it to return -1 for not being on a wall yet, and just had logic elsewhere check if timeToBurnout == 0 to see if it's burnt out since the time for a flare on wall doesn't seem to be able to go below 0 anyways (compares the time with 0 and chooses the max before returning).

### DIFF
--- a/core/assets/jsons/global.json
+++ b/core/assets/jsons/global.json
@@ -180,6 +180,7 @@
     "radius":      	  0.1,
     "lightradius":     3.5,
     "flareduration":  2500,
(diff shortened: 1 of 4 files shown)


BryanGraeser (approved):
Looks good to me

PR #173, opened 2020-05-21
Story modes

lsizemore8:
Added the three story modes. In the game, here's where they are: 
intro: When you choose level 1 of cave
Trees: Finish the last level of cave
Volcano: finish the last level of trees

Currently the story will always play after you complete that x task. It's designed to still work, even as we add/remove levels (provided there is 1 level in the first world). Music is the menu music for now

This also fixes the fog black path for patrollers, as it was a 2 line change and not worth making a separate PR on


stallop (commented):
Moved around levels a bit and wasn't able to break it, so functionally it seems all good. Only issue is that the left and right buttons seem a bit hard to see on some of the brighter comics. Maybe a slight black outline would help? Or just anything to visually distinguish them from the background a bit more.

BryanGraeser:
Looks good just needs the new buttons

BryanGraeser:
Question: Is there a way to re-view these?

BryanGraeser:
Maybe add them to the map?

stallop (approved):
Much easier to see now.

PR #174, opened 2020-05-21
Check framerate

ronnie-dumesh:
This code adapts our texture updating by being dynamic for any framerate (although primarily 30 FPS and 60 FPS were tested).

It also adds a fireBuddyLimit to global.json so that the updating of the fire buddy does not relay on player movement at all. The fire buddy now "moves" even when the player idle. 

Additionally, it makes minor modifcations to the speed of the victory animation.

Essentially, the game should "move" at the same-rate regardless of the FPS. You can change the FPS by uncommenting out the code in DesktopLauncher. It may look stuttery due to the low FPS, but all animations should move at a constant rate and shouldn't look like "slow-motion"

### DIFF
--- a/core/assets/jsons/global.json
+++ b/core/assets/jsons/global.json
@@ -26,7 +26,7 @@
       "death-right": "player-death-right"
     },
     "deathdelay": 45,
-    "windelay": 45,
+    "windelay": 20,
     "walksound":    "new-walk",
     "firebuddy": {
       "texture": {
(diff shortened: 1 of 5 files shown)


lsizemore8 (approved):
Remove the commented code but this looks gtg

BryanGraeser:
If this PR is meant to speed up the win animation, I think it can still be significantly faster. At least double what it is rn

BryanGraeser:
Other than that looks good

ronnie-dumesh:
I made it as fast as I possibly could for it to work for 30 FPS and 60 FPS, a duration of about 1 second

PR #176, opened 2020-05-22
fixed player walking upwards using walkRight filmstrip

ronnie-dumesh:


core/src/com/fallenflame/game/CharacterModel.java
```diff
@@ -330,10 +330,10 @@ public void update(float dt) {
         if(angle < 0) angle = angle + 2 * Math.PI;
         int angle100 = (int) (angle * 100);
 
-        if(angle100 == 0){
+        if(angle100 == 0 || angle100 == 628){
```
lsizemore8, line 333: Why 628? This seems like a random number

ronnie-dumesh: 2 * pi  aka 0 on a unit circle. I convert 6.28 to an int just to prevent floating point errors in comparisons and stuff

lsizemore8 (approved):
Looks gtg

PR #177, opened 2020-05-22
refactor light radius code and fix glitch that would allow sprint rad…

ronnie-dumesh:
…ius without sprinting

Previously, it was possible to press shift and ctrl together and then let go of ctrl that made a permenant light radius the size of the sprint radius

### DIFF
--- a/core/src/com/fallenflame/game/GameEngine.java
+++ b/core/src/com/fallenflame/game/GameEngine.java
@@ -778,10 +778,6 @@ public class GameEngine implements Screen, InputProcessor {
      * @param amount representing if the wheel scrolled down (1) or up (-1). Can only be those two values.
      * @return boolean saying if the event was handled*/
     public boolean scrolled (int amount) {
-        if(!isScreenActive()){
-            return true;
-        }
-
         return true;
     }
 
--- a/core/src/com/fallenflame/game/GameEngine.java
+++ b/core/src/com/fallenflame/game/GameEngine.java
@@ -692,15 +692,6 @@ public class GameEngine implements Screen, InputProcessor {
         if (Gdx.input.isKeyPressed(InputBindings.getBindingOf(InputBindings.Control.GO_DOWN))) {
             vertical -= 1.0f;
         }
-
-        //#region mouse wheel alternative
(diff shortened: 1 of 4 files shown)


core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -790,12 +781,6 @@ public boolean scrolled (int amount) {
         if(!isScreenActive()){
             return true;
         }
-        if(amount == 1 && ALLOW_DEBUG){
```
lsizemore8, line 793: Why is this causing the bug/ This should never be true unless we have ALLOW_DEBUG, as the light radius was a useful debugger

core/src/com/fallenflame/game/LevelController.java
```diff
@@ -640,12 +640,13 @@ public void update(float dt) {
...
-        // If dying or dead, that's it. Don't update anything else.
+        // If player won or is dead, that's it. Don't update anything else.
         // (such as light, fog, enemies, etc)
-        if (player.isDead() || player.isDying() || player.hasWon()) return;
+        if (player.isDead() || player.hasWon()) return;
 
-        if (player.isWinning()) {
-            player.setLightRadius(player.getLightRadiusSprint()); //increase light radius to see fire buddy
+        //Update nothing except lights
+        if (player.isWinning() || player.isDying()) {
```
lsizemore8, line 648: If this happens when they're dying, you've now introduced a huge benefit to them, as they can see more as they die/get a lot more of their surrounding than intended

lsizemore8 (changes requested):
Light radius changes look good, dying going to max light radius does not (try level 8, you give away a whole lot w/ the max light radius with death)

core/src/com/fallenflame/game/GameEngine.java
```diff
@@ -790,12 +781,6 @@ public boolean scrolled (int amount) {
         if(!isScreenActive()){
             return true;
         }
```
lsizemore8: Just return true

lsizemore8 (approved):
Looks good

PR #178, opened 2020-05-22
Sound balance

stallop:
-Made it so constant noise doesn't loop. Instead has a 1/200 chance of playing on any frame, and then can't play until 125 more frames have passed (makes sure one enemy can't play the sound multiple times at once). Figured since we have fog now, sound isn't always necessary to tell where enemies in darkness are and should instead be used sparingly.
-Made it so enemies can't be heard unless you're decently close to them using the threshold value.
-Trimmed all the mp3 files so they don't have as much silence at the end. Still kept a little bit b/c it sounded weird with absolutely no pause before restarting.


BryanGraeser:
I'm not sure what the 1/200 purpose is, but I feel like this feels very inconsistent from a player's perspective. Sometimes the enemies make sounds, sometimes they don't with no real player-visible reason as to why. It also feels very weird for an enemy chasing you to make no noise, but other non-active enemies to be making noise (I think regardless of what we do enemies chasing you should be audible). I also found that when I played many enemies did not make any noise. 

Idk, in short it just feels too arbitrary. Sometimes everything sounds and plays good, sometimes I hear nothing. I would also check the background music volume (if you haven't)

BryanGraeser:
I like this better. Last thing: when an enemy chases you, often they chase at a distance where you can't hear them. This feels weird. Any way we can have chasing enemies' have a wider audio range?

PR #179, opened 2020-05-22
Sound balance

stallop:
move sound has a little more range


lsizemore8:
I think it's better. Maybe a little bit more, so it's a tad more of "something's always there", but it's good if not close

BryanGraeser (approved):
gtg

PR #181, opened 2020-05-22
page bug fixed

stallop:
I believe I fixed it, hopefully didn't break anything else in the process.


lsizemore8 (changes requested):
This breaks it. When you press "next", it now goes to the "last page", which is an empty page. When you press previous, it then shows the levels 11-16, and you can press the prev button again to get back to 1-10

lsizemore8 (approved):
Seems to work now

PR #185, opened 2020-05-22
Add ronnie

lsizemore8:
Added Ronnie's level and removed Max's, as the tight corridors in it made it impossible to throw flares accurately, making it less fun.


foxhatleo (approved):
Preapproved.

